PR #35, opened 2026-09-06
Leave pushing and deciding a pull request to Bertan

### LINKED ISSUE #37: Finish the boundary's own claims before PR #35 merges
## Parent

#36

## What to build

PR #35 introduces the agent boundary, and four of the things it says about
itself are wrong, missing, or unenforceable. This ticket makes the branch's
claims true before it lands.

An agent's worktree is created from the active dev branch rather than from
`main`, so that a worktree branch contains the work it is proposed alongside.
CLAUDE.md stops asserting that the boundary can rest on nothing else, and says
instead that it rests on these hooks *because* the agent shares Bertan's
credentials — a choice open to revision, not a law. The fact that nothing guards
`.claude/` is named as deliberately open rather than left for a reader to
discover. The hooks gain a written stopping rule, so the next person to find an
evasion has a standard for deciding whether it is worth fixing. And the argument
for why any of this is enforced by hooks at all moves out of CLAUDE.md into the
repository's first ADR, leaving CLAUDE.md with the paragraph an agent actually
needs each session.

The ADR records the mechanism, not the inventory: *why hooks rather than a
server-side ruleset*, with the rejected alternative — giving the agent its own
actor through a deploy key or an installation token — written down because the
rejection is the part a future reader would otherwise re-propose blind. Written
that way, it stays true when ticket 4 widens what the boundary refuses.

## Acceptance criteria

- [ ] `worktree.baseRef` is set in the checked-in settings so a worktree is
      created from the active dev branch; verified by creating one and
      confirming its fork point is the active dev branch and not `main`
- [ ] Both hook files carry the stopping rule: a newly found evasion earns a fix
      only if it is a shape an agent would plausibly write, not one it would
      have to construct
- [ ] CLAUDE.md no longer claims the boundary "can rest on nothing else", and
      states the shared-credentials reason as a choice rather than a constraint
- [ ] The "deliberately left open" section names that nothing guards `.claude/`,
      together with the standard that excludes it
- [ ] CLAUDE.md's boundary section is reduced to its operative paragraph
- [ ] `docs/adr/0001` exists, states the principle and the rejected alternative,
      and contains no inventory of permitted commands
- [ ] The check suite passes, with the same count as before this ticket
- [ ] Everything lands on the branch of PR #35

## Blocked by

None (can start immediately).

bgunyel:
Closes #37

Two `PreToolUse` hooks on the `Bash` matcher, joining the four already there.

**`no-git-push.sh`** refuses every push. Pushes are made by Bertan from a separate terminal outside Claude Code, so no push is judged case by case.

**`no-pr-decisions.sh`** refuses merging a pull request, reviewing one with a verdict (`--approve`, `--request-changes`), closing or reopening one, and creating or deleting a release. The line is between talking and deciding: an agent may open a pull request and argue on it, and may not accept, reject or merge one.

Still allowed: `gh pr create`, `gh pr comment`, `gh pr edit`, `gh pr view/list/diff/checks/ready`, `gh pr review --comment`, and every `gh issue` subcommand.

### The direct API path is matched too

`gh pr merge` is only the convenient spelling. The same merge is one REST call (`PUT /repos/O/R/pulls/N/merge`) or one graphql mutation away, and `gh api` is allowlisted in `settings.local.json`, so it runs without a prompt.

This section used to claim more than the code did — it covered merging and reviewing, and left closing and releasing open through `gh api` while refusing both one file-section earlier. Raised in review; the code was widened rather than the claim narrowed. What is matched now, and it is a list rather than a principle:

| decision | REST | graphql |
|---|---|---|
| merge | `/pulls/*/merge` | `mergePullRequest` |
| review with a verdict | `/pulls/*/reviews` | `addPullRequestReview` |
| close / reopen | a write to `/pulls/N` carrying `state=closed` or `state=open` | `closePullRequest`, `reopenPullRequest`, `updatePullRequest` with that state |
| release | `/releases` | `createRelease`, `updateRelease`, `deleteRelease` |

Close and reopen are matched on the **field**, not the endpoint, because `PATCH /repos/O/R/pulls/N` is also how `gh pr edit` retitles a PR — which is allowed, and is asserted to stay allowed by a probe.

**Only writes are matched.** `gh` sends `GET` unless a `--method` or a field flag says otherwise, and the endpoint does not distinguish a decision from a reading: `GET /pulls/N/reviews` lists reviews and `GET /pulls/N/merge` reports whether the PR is merged. Both are reading a pull request, which the boundary allows. A method that cannot be parsed counts as a write. The exception is a call wrapped in `bash -c`, where the method is quoted text and no more readable than the endpoint — a wrapped read is refused with the writes.

### Two rounds of false positives, both recorded as probes

Matching is anchored to a *command position* — start of line, or after `;` `&&` `||` `|` or an opening paren — not to a bare preceding space, which is what `no-commit-to-main.sh` uses. The looser anchor blocks writing a commit message that merely names the command, and this repository does that constantly.

That was not enough. `grep` anchors `^` per line, so a wrapped line of prose inside a quoted heredoc still read as a command position — **the first version of these hooks refused the commit that introduced them.** Heredoc bodies are now dropped before matching. Because that also hides a real command, shell wrappers (`sh -c`, and a heredoc fed to a shell) are matched separately on the raw text, quotes and bodies included.

### `probe-hooks.sh`

63 probes, run with `bash .claude/hooks/probe-hooks.sh`, all passing:

- 4 that a commit message or note may *name* these commands
- 4 that a real command in multi-line input is still caught
- 55 over the ordinary spellings, both the block and the allow side

It found two defects during this work, which is why it is committed rather than thrown away. A hook is a process and running it is the only way to test it; per CLAUDE.md the expectations are written as literals, never derived from the scripts.

**Reviewer's call:** this third file is the one thing here that was not asked for. Drop it if you would rather the hooks ship alone.

### Notes

`no-commit-to-main.sh` already refused pushes whose destination was main, and that rule is now a subset of the broader one. Left in place on purpose: it carries the narrower message, and it still stands if the broader hook is disabled.

These hooks are inert in the main checkout until this merges — `settings.json` there was restored to its committed state during this work.

**These stop mistakes, not adversaries.** A caller who means to push or merge can spell it in ways no regex catches.

https://claude.ai/code/session_01Mx8Qck1kMHWmNFhKb4YAjv

### DIFF
--- a/.claude/hooks/lib/command-scan.sh
+++ b/.claude/hooks/lib/command-scan.sh
@@ -1,0 +1,126 @@
+#!/bin/bash
+# Where a command starts, where its arguments end, and how many commands a
+# string holds. Sourced by no-git-push.sh and no-pr-decisions.sh.
+#
+# This exists because of what the defects in PR #35 turned out to have in
+# common. Every one of them, found by Bertan or by the assistant, was the same
+# question answered differently in a different place -- how far around a matched
+# token to look:
+#
+#   - the anchor required column zero, so an indented command was not a command
+#   - wrapper re-admission sufficed for a heredoc and did nothing for sh -c
+#   - the option scope was the whole line, so `ls --all && git push` read as --all
+#   - narrowing it to one command cut at a newline, so a backslash continuation
+#     made every push look bare, which is the permitted shape
+#   - and it found the first command only, so a second one after ; or && was
+#     never examined at all
+#
+# Two of those were introduced by the fix to the previous two. The rule was
+# re-derived in about a dozen regular expressions across two files, so fixing it
+# in one place kept opening it in another. It is derived once here instead, and
+# the probe suite points at these functions directly.
+#
+# The answers are approximate on purpose. Splitting more eagerly than a shell
+# would yields extra command candidates, which can only refuse more; it never
+# hides one. That is the safe direction for a guard whose failure mode, twice
+# now, has been to report the permitted answer.
+
+# Reduce a raw command to lines that can be scanned: heredoc bodies dropped,
+# line continuations joined.
+#
+# A heredoc body is data, not commands. This repository writes dev-log entries
+# and commit messages through a quoted heredoc, and those texts name the very
+# commands the hooks refuse; grep anchors ^ per line, so a line of prose
+# beginning with one read as a command position, and an early version of the
+# hooks refused the commit that introduced them.
+#
+# Joining runs after dropping, so a backslash at the end of the line before a
+# heredoc terminator cannot swallow the terminator and hide what follows.
+cs_normalise() {
+  awk '
+    ind { if ($0 == d) ind = 0; next }
+    {
+      if (match($0, /<<-?[[:space:]]*[^[:space:];|&<>()]+/)) {
+        d = substr($0, RSTART, RLENGTH)
+        sub(/^<<-?[[:space:]]*/, "", d)
+        gsub(/[\047"]/, "", d)
+        ind = 1
+      }
+      print
+    }' \
+  | awk '
+    {
+      line = $0
+      while (line ~ /\\$/) {
+        sub(/\\$/, "", line)
+        if ((getline nxt) > 0) line = line nxt; else break
+      }
+      print line
+    }'
+}
+
+# Print one command per line, with anything that precedes the command word
+# removed, so a caller matches on ^ and never has to describe a command
+# position again.
+#
+# Separators are ; && || | ( ) and a backtick. The backtick is there because
+# $( ) was closed by the paren and its twin was not -- the same asymmetry
+# GIT_DIR= had against --git-dir.
+#
+# Removed prefixes: environment assignments, and the wrapper words that run
+# another command with their own options. A caller that cares about the
+# assignments themselves must look at the un-split text; no-git-push.sh does,
+# for GIT_DIR= and GIT_WORK_TREE=.
+cs_split() {
+  sed -e 's/&&/\n/g' -e 's/||/\n/g' -e 's/[;&|()`]/\n/g' \
+  | awk '
+    {
+      line = $0
+      sub(/^[[:space:]]+/, "", line)
+      changed = 1
+      while (changed) {
+        changed = 0
+        if (match(line, /^[A-Za-z_][A-Za-z0-9_]*=[^[:space:]]*[[:space:]]+/)) {
+          line = substr(line, RSTART + RLENGTH)
+          changed = 1
+        }
+        if (match(line, /^(env|command|xargs|nohup|nice|time|stdbuf|ionice)[[:space:]]+/)) {
+          line = substr(line, RSTART + RLENGTH)
+          while (match(line, /^-[^[:space:]]*[[:space:]]+/)) {
+            line = substr(line, RSTART + RLENGTH)
+          }
+          changed = 1
+        }
+      }
+      sub(/[[:space:]]+$/, "", line)
+      if (line != "") print line
+    }'
+}
+
+# Print the arguments of a git subcommand and succeed, or print nothing and fail
+# if this command is not `git <subcommand>`. Global options are skipped,
+# including the two that take a separate value: without that, -C /path ends the
+# match before the subcommand is reached.
+#
+# The exit status is what distinguishes `git push` -- a push whose argument list
+# is empty, and the permitted shape -- from a command that is not a push at all.
+# A caller testing the printed text instead would have to re-derive the rule,
+# which is the habit this file exists to end.
+cs_git_args() {
+  awk -v want="$1" '
+    BEGIN { found = 0 }
+    {
+      line = $0
+      if (line !~ /^git([[:space:]]|$)/) next
+      sub(/^git[[:space:]]*/, "", line)
+      while (match(line, /^(-[cC][[:space:]]+[^[:space:]]+|--(git-dir|work-tree|namespace|exec-path)=[^[:space:]]*|-[^[:space:]]+)[[:space:]]+/)) {
+        line = substr(line, RSTART + RLENGTH)
+      }
+      if (line !~ "^" want "([[:space:]]|$)") next
+      sub("^" want "[[:space:]]*", "", line)
+      print line
+      found = 1
+      exit
+    }
+    END { exit(found ? 0 : 1) }'
+}
--- a/.claude/hooks/lib/command-scan.sh
+++ b/.claude/hooks/lib/command-scan.sh
@@ -28,6 +28,14 @@
 # after `then`, `do`, `else`, `{` or `!` is at a command position and was not
 # treated as one. Each of the three hid every command that followed it.
 #
+# A third review found the heredoc question wrong a third time, and that is the
+# number that settled it. `git commit -m "fix <<EOF handling"` contains no
+# heredoc -- inside double quotes `<<` is text -- and the rest of the command
+# was dropped. Three wrong answers, each silent and each in the permitting
+# direction, is evidence about the question rather than about the answers: it
+# cannot be got exact by looking more carefully, because that is what the
+# previous two attempts were. The drop is a fail-safe now. See cs_normalise.
+#
 # The answers are approximate on purpose. Splitting more eagerly than a shell
 # would yields extra command candidates, which can only refuse more; it never
 # hides one. That is the safe direction for a guard whose failure mode, twice
@@ -53,12 +61,26 @@
 # tabs from the terminator, which an exact comparison never matched, so that
 # heredoc did not end either. Either one turned a `git push --mirror` or a
 # `gh pr merge` on a following line into nothing at all.
+#
+# Those were the second and third answers to the same question, and a fourth
+# followed them: `git commit -m "fix <<EOF handling"` has no heredoc in it at
+# all -- inside double quotes `<<` is text -- and the opener was matched
+# anywhere on the line, quotes included. So the drop is no longer trusted to be
+# right. A heredoc that never reaches its terminator was not a heredoc, and the
+# lines held for it are given back at END rather than lost.
+#
+# That is the answer this question should have had from the start: the exact
+# version has been got wrong three times, and each time the failure was silent
+# and in the permitting direction. The fail-safe costs a genuinely unterminated
+# heredoc being scanned as commands -- which bash would refuse to run anyway --
+# and it is the direction this file takes everywhere else.
 cs_normalise() {
   awk '
     ind {
       line = $0
       if (dash) sub(/^\t+/, "", line)
-      if (line == d) ind = 0
+      held[++nheld] = $0
+      if (line == d) { ind = 0; nheld = 0 }
       next
     }
     {
@@ -74,7 +96,10 @@ cs_normalise() {
         ind = 1
       }
       print
-    }' \
+    }
+    # The terminator never arrived, so this was not a heredoc and the lines were
+    # dropped in error. Give them back.
+    END { for (i = 1; i <= nheld; i++) print held[i] }' \
   | awk '
     {
       line = $0
--- a/.claude/hooks/lib/command-scan.sh
+++ b/.claude/hooks/lib/command-scan.sh
@@ -20,6 +20,14 @@
 # in one place kept opening it in another. It is derived once here instead, and
 # the probe suite points at these functions directly.
 #
+# A second review found three more of the same shape, and all three were here
+# rather than spread across the hooks, which is the point of the file. Two were
+# the heredoc question answered too loosely -- a here-string read as a heredoc,
+# and a tab-indented <<- terminator never recognised, so that heredoc never
+# ended. One was the command-position question answered too narrowly: a command
+# after `then`, `do`, `else`, `{` or `!` is at a command position and was not
+# treated as one. Each of the three hid every command that followed it.
+#
 # The answers are approximate on purpose. Splitting more eagerly than a shell
 # would yields extra command candidates, which can only refuse more; it never
 # hides one. That is the safe direction for a guard whose failure mode, twice
@@ -36,12 +44,31 @@
 #
 # Joining runs after dropping, so a backslash at the end of the line before a
 # heredoc terminator cannot swallow the terminator and hide what follows.
+#
+# Dropping is the one step here that hides commands rather than exposing them,
+# so what counts as a heredoc has to be exact in both directions -- and it was
+# wrong in both. `<<<` is a here-string: the operator regex matched its second
+# and third `<`, took the here-string's own text for a terminator that never
+# arrives, and dropped the rest of the command. `<<-` lets bash strip leading
+# tabs from the terminator, which an exact comparison never matched, so that
+# heredoc did not end either. Either one turned a `git push --mirror` or a
+# `gh pr merge` on a following line into nothing at all.
 cs_normalise() {
   awk '
-    ind { if ($0 == d) ind = 0; next }
+    ind {
+      line = $0
+      if (dash) sub(/^\t+/, "", line)
+      if (line == d) ind = 0
+      next
+    }
     {
-      if (match($0, /<<-?[[:space:]]*[^[:space:];|&<>()]+/)) {
-        d = substr($0, RSTART, RLENGTH)
+      # A here-string is not a heredoc. Blanked at its own width, so a real
+      # heredoc later on the same line is still found where it stands.
+      scan = $0
+      gsub(/<<</, "   ", scan)
+      if (match(scan, /<<-?[[:space:]]*[^[:space:];|&<>()]+/)) {
+        d = substr(scan, RSTART, RLENGTH)
+        dash = (d ~ /^<<-/)
         sub(/^<<-?[[:space:]]*/, "", d)
         gsub(/[\047"]/, "", d)
         ind = 1
@@ -67,10 +94,22 @@ cs_normalise() {
 # $( ) was closed by the paren and its twin was not -- the same asymmetry
 # GIT_DIR= had against --git-dir.
 #
-# Removed prefixes: environment assignments, and the wrapper words that run
-# another command with their own options. A caller that cares about the
-# assignments themselves must look at the un-split text; no-git-push.sh does,
-# for GIT_DIR= and GIT_WORK_TREE=.
+# Removed prefixes: environment assignments, the shell's own control words, and
+# the wrapper words that run another command with their own options. A caller
+# that cares about the assignments themselves must look at the un-split text;
+# no-git-push.sh does, for GIT_DIR= and GIT_WORK_TREE=.
+#
+# The control words are here because a separator is not the only thing a command
+# can follow. `if true; then git push --mirror origin; fi` splits correctly and
+# still left `then` standing in front of the command word, so the anchor never
+# saw a push at all; `do`, `else`, `elif`, `{` and `!` each did the same. They
+# are removed rather than matched around, so every caller keeps anchoring at ^.
+#
+# The trade, taken knowingly and probed as such: a quoted string holding a
+# separator and then one of these words in front of a refused command now reads
+# as that command, so `git commit -m "wait; then git push --all origin"` is
+# refused. That is the direction this file has taken throughout -- a blocked
+# comment is visible and one edit away, a silently permitted push is neither.
 cs_split() {
   sed -e 's/&&/\n/g' -e 's/||/\n/g' -e 's/[;&|()`]/\n/g' \
   | awk '
@@ -81,6 +120,10 @@ cs_split() {
       while (changed) {
         changed = 0
         if (match(line, /^[A-Za-z_][A-Za-z0-9_]*=[^[:space:]]*[[:space:]]+/)) {
+          line = substr(line, RSTART + RLENGTH)
+          changed = 1
+        }
+        if (match(line, /^([{}!]|if|then|elif|else|fi|while|until|for|do|done|case|esac|select|function|coproc)([[:space:]]+|$)/)) {
           line = substr(line, RSTART + RLENGTH)
           changed = 1
         }
--- a/.claude/hooks/no-git-push.sh
+++ b/.claude/hooks/no-git-push.sh
@@ -19,96 +19,59 @@
 # convenience in front of one. main is different -- there the policy is the same
 # for both, and the branch ruleset is the right mechanism.
 #
+# Finding commands in the text is lib/command-scan.sh's job, not this file's.
+# Every defect found in PR #35 was that question answered differently in a
+# different place; it is answered once there now.
+#
 # no-commit-to-main.sh independently refuses pushes whose destination is main.
 # It is left in place: it carries the narrower message, and it still stands if
 # this file is disabled.
 #
 # This stops mistakes, not adversaries.
+. "$(dirname "$0")/lib/command-scan.sh"
+
 INPUT=$(cat)
 COMMAND=$(echo "$INPUT" | jq -r '.tool_input.command')
+SCAN=$(printf '%s\n' "$COMMAND" | cs_normalise)
+CMDS=$(printf '%s\n' "$SCAN" | cs_split)
 
-# A heredoc body is data, not commands. This repository writes dev-log entries
-# and commit messages through a quoted heredoc, and those texts name the very
-# commands refused here. grep anchors ^ per line, so a line of prose beginning
-# with one of them read as a command position, and an earlier version of this
-# hook refused the commit that introduced it. Bodies are dropped before
-# matching.
-SCAN=$(echo "$COMMAND" | awk '
-  ind { if ($0 == d) ind = 0; next }
-  {
-    if (match($0, /<<-?[[:space:]]*[^[:space:];|&<>()]+/)) {
-      d = substr($0, RSTART, RLENGTH)
-      sub(/^<<-?[[:space:]]*/, "", d)
-      gsub(/[\047"]/, "", d)
-      ind = 1
-    }
-    print
-  }')
-
-# A backslash-newline is a line continuation, not a command separator. The
-# argument scope further down runs from push to the next separator, a newline
-# ended it, and an empty scope fell through to the bare-push case -- the
-# permitted one. So `git push \` followed by `--mirror origin` read as an
-# ordinary bare push and would have deleted every remote branch absent locally.
-# Continuations are joined before anything is matched. Reported on PR #35.
-SCAN=$(printf '%s\n' "$SCAN" | awk '
-  {
-    line = $0
-    while (line ~ /\\$/) {
-      sub(/\\$/, "", line)
-      if ((getline nxt) > 0) line = line nxt; else break
-    }
-    print line
-  }')
-
-# Dropping bodies is itself a bypass: `sh -c "..."` and a heredoc fed to a shell
-# both carry real commands inside text that was just discarded.
-WRAPPED=
-if echo "$COMMAND" | grep -qE '(^[[:space:]]*|[;&|(][[:space:]]*)([A-Za-z_][A-Za-z0-9_]*=[^[:space:]]*[[:space:]]+)*((ba|z|)sh[[:space:]]+(-c|<<)|eval([^-A-Za-z0-9_]|$))'; then
-  WRAPPED=1
-fi
+REFUSE="Blocked: git push. An agent may push only the branch of the linked worktree it is working in, so that it can open a pull request."
 
 # A push inside a wrapper is refused outright, with no worktree exception. The
 # payload sits in quotes where no command position exists, so the destination
 # cannot be read; permitting a push whose target is unknown is not the same as
-# permitting this branch.
-if [ -n "$WRAPPED" ] \
+# permitting this branch. Matched on the raw command, quotes and heredoc bodies
+# included, because that is where the payload still is.
+#
+# This runs before asking whether there is a push, and must: a wrapped push has
+# no command word for the tokeniser to find, so the question would answer no and
+# the hook would leave. Ordering it the other way let all four wrapper forms
+# through, which the probe suite caught.
+if echo "$COMMAND" | grep -qE '(^[[:space:]]*|[;&|(`][[:space:]]*)([A-Za-z_][A-Za-z0-9_]*=[^[:space:]]*[[:space:]]+)*((ba|z|)sh[[:space:]]+(-c|<<)|eval([^-A-Za-z0-9_]|$))' \
    && echo "$COMMAND" | grep -qE 'git[[:space:]]+([^;&|]*[[:space:]])?push([^-A-Za-z0-9_]|$)'; then
   echo "Blocked: git push inside a shell wrapper. The destination cannot be read through a quoted payload, so the worktree exception does not apply. Push plainly from the worktree, or leave it to Bertan." >&2
   exit 2
 fi
 
-# A command sits at a command position: the start of a line -- leading
-# whitespace included -- or after ; && || | or an opening paren.
-#
-# Leading whitespace is load-bearing. The anchor first written here required
-# column zero, which fixed the heredoc false positive and silently gave up every
-# indented push: a push inside an if or a for loop is written indented, and that
-# is the accident this hook exists to stop. Reported on PR #35 and fixed there.
-#
-# The cost is one false positive, accepted knowingly: a quoted multi-line string
-# that is not a heredoc, whose continuation line begins with the command. A
-# blocked comment is visible and one edit away; a permitted push is neither.
-#
-# The -c/-C branch lets `git -C /path push` through the global-option run;
-# without it the option's separate value ends the match before push is reached.
-if ! echo "$SCAN" | grep -qE '(^[[:space:]]*|[;&|(][[:space:]]*)([A-Za-z_][A-Za-z0-9_]*=[^[:space:]]*[[:space:]]+)*git[[:space:]]+((-[cC][[:space:]]+[^ ]+|-[^ ]+)[[:space:]]+)*push([^-A-Za-z0-9_]|$)'; then
-  exit 0
-fi
-
-REFUSE="Blocked: git push. An agent may push only the branch of the linked worktree it is working in, so that it can open a pull request."
+# Is there a push here at all? A command with no push in it costs one pass and
+# leaves without an opinion.
+HAVE_PUSH=
+while IFS= read -r CMD; do
+  if cs_git_args push <<<"$CMD" >/dev/null; then HAVE_PUSH=1; break; fi
+done <<CMDLIST
+$CMDS
+CMDLIST
+[ -n "$HAVE_PUSH" ] || exit 0
 
 # Where the hook runs is the session's directory, which is not necessarily where
 # the command will run. Anything that moves the push elsewhere is refused rather
-# than assessed: cd and pushd, and git redirected by option or by environment.
-# GIT_DIR= and GIT_WORK_TREE= are the environment spelling of --git-dir and
-# --work-tree and were missed when only the option form was refused.
-if echo "$SCAN" | grep -qE '(^[[:space:]]*|[;&|(][[:space:]]*)([A-Za-z_][A-Za-z0-9_]*=[^[:space:]]*[[:space:]]+)*(cd|pushd|popd)([^-A-Za-z0-9_]|$)'; then
+# than assessed. The environment spellings are read from the un-split text,
+# because cs_split removes assignments to find the command word behind them.
+if printf '%s\n' "$CMDS" | grep -qE '^(cd|pushd|popd)([^-A-Za-z0-9_]|$)'; then
   echo "$REFUSE This command changes directory first, so where the push would land cannot be judged from here." >&2
   exit 2
 fi
-if echo "$SCAN" | grep -qE '(GIT_DIR|GIT_WORK_TREE|GIT_COMMON_DIR)=' \
-   || echo "$SCAN" | grep -qE 'git[[:space:]]+([^;&|]*[[:space:]])?(-C|--git-dir|--work-tree)([[:space:]]|=)'; then
+if echo "$SCAN" | grep -qE '(GIT_DIR|GIT_WORK_TREE|GIT_COMMON_DIR)='; then
   echo "$REFUSE This command points git at another directory, so where the push would land cannot be judged from here." >&2
   exit 2
 fi
@@ -130,56 +93,6 @@ if [ "$CURRENT" = "main" ] || echo "$CURRENT" | grep -qE '^dev-[0-9]+$'; then
   exit 2
 fi
 
-# Everything after the first push on the matched line, up to the next shell
-# separator, is the push's own arguments. Every check below reads these rather
-# than the whole command, so an unrelated flag elsewhere on the line -- the -f
-# of `rm -f x && git push` -- is not mistaken for the push's own.
-PUSH_LINE=$(echo "$SCAN" | grep -m1 -E '(^[[:space:]]*|[;&|(][[:space:]]*)([A-Za-z_][A-Za-z0-9_]*=[^[:space:]]*[[:space:]]+)*git[[:space:]]+((-[cC][[:space:]]+[^ ]+|-[^ ]+)[[:space:]]+)*push([^-A-Za-z0-9_]|$)')
-ARGS=${PUSH_LINE#*push}
-# The scope ends at a shell separator, and a closing paren is one: without it
-# the ) of `(git push)` was read as the push's remote.
-ARGS=$(printf '%s' "$ARGS" | sed 's/[;&|)].*//' | tr -d '\042\047')
-
-# Joining above consumes a backslash that ends a line, including one with no
-# continuation line after it, which is simply a bare push. A backslash surviving
-# here is one the join did not recognise -- trailed by whitespace, say -- and
-# what follows it cannot be read. An empty scope is the permitted case, so an
-# unreadable one is refused rather than allowed to look empty.
-if printf '%s' "$ARGS" | grep -q '\\[[:space:]]*$'; then
-  echo "$REFUSE The arguments continue past where this check can read them." >&2
-  exit 2
-fi
-
-# Whole-repository and tag forms name no branch at all. Refusing branches by
-# name was the shape of this check before, and a denylist cannot see a spelling
-# that names nothing: `git push --all origin` advanced main and dev-05 from any
-# worktree, and --mirror deleted every remote branch absent locally, closing
-# open pull requests. Reported on PR #35.
-if printf ' %s ' "$ARGS" | grep -qE '[[:space:]](--all|--mirror|--tags|--follow-tags|--prune|--delete|-d)([[:space:]]|=|$)'; then
-  echo "$REFUSE That form pushes or deletes refs wholesale rather than naming this branch." >&2
-  exit 2
-fi
-
-# A forced push rewrites what the remote already has, which for this branch is
-# the history an open pull request is showing. --force-with-lease is refused
-# with the rest: it makes the rewrite safe against clobbering someone else's
-# work, not against rewriting a PR under its reviewer. The short form may be
-# bundled with other single-letter options, as -fu, so any single-dash token
-# containing f counts.
-if printf ' %s ' "$ARGS" | grep -qE '[[:space:]](--force|--force-with-lease|--force-if-includes|-[A-Za-z]*f[A-Za-z]*)([[:space:]]|=|$)'; then
-  echo "$REFUSE That is a forced push, which rewrites history the open pull request is showing. Add a commit instead." >&2
-  exit 2
-fi
-
-if printf '%s' "$ARGS" | grep -q '[*]'; then
-  echo "$REFUSE A wildcard refspec does not name this branch." >&2
-  exit 2
-fi
-
-# What remains must positively name this branch: the first bare token is the
-# remote and every later one is a refspec, whose source and destination must
-# both be this branch or HEAD.
-
 names_this_branch() {
   case "$1" in
     HEAD|"$CURRENT"|"refs/heads/$CURRENT") return 0 ;;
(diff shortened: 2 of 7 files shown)


bgunyel:
Ran both scripts from this branch against `stdin` shaped like the real hook payload. Two defects, both inside the PR's own threat model.

## 1. Indentation defeats both hooks

The anchor is `(^|[;&|(]\s*)`. `^` requires the command at column zero, and no separator class includes whitespace, so anything indented passes:

```
ALLOW   if true; then\n    git push origin dev-05\nfi
ALLOW   for b in a b; do\n  git push origin $b\ndone
ALLOW   if true; then\n    gh pr merge 35\nfi
```

This is not an adversarial spelling — it is how a conditional or a loop is written — which makes it the accident these hooks exist to stop.

**`no-commit-to-main.sh` does not have this bug.** With the bare-space anchor described here as the looser version, it blocks the indented push to main:

```
BLOCK   if true; then\n    git push origin main\nfi     (no-commit-to-main.sh)
ALLOW   if true; then\n    git push origin dev-05\nfi   (no-git-push.sh)
```

The tighter anchor fixed the heredoc false positive and traded away a true positive. That trade is not recorded in the PR description, which presents the change as a pure improvement.

## 2. `no-pr-decisions.sh` has no shell-wrapper rule

`no-git-push.sh` ends with a rule that re-matches the raw command for `sh -c`, `eval` and a heredoc fed to a shell. `no-pr-decisions.sh` drops heredoc bodies but has no equivalent fallback, so everything the dropping hides stays hidden:

```
BLOCK   gh pr merge 35
ALLOW   bash -c 'gh pr merge 35'
ALLOW   sh -c "gh pr merge 35"
ALLOW   eval 'gh pr merge 35'
ALLOW   gh api graphql -f query=@- <<EOF ... mergePullRequest ... EOF
```

The last one bears directly on *"The direct API path is matched too"*: a heredoc is the ordinary way to send a graphql mutation, and the body is dropped before matching, so the mutation is never seen.

Two files written in one commit, one guarded against wrappers and one not, reads as an oversight rather than a decision.

## On the fixes

The wrapper rule is a copy of the sibling's, matched on the raw `$COMMAND`.

The anchor is **not** a free change. Widening to `(^[[:space:]]*|[;&|(]\s*)` is safe against heredoc prose, since bodies are already dropped — but it reintroduces the false positive for a **quoted multi-line string that is not a heredoc**, e.g. `gh issue comment -b "...\n  git push ..."`. Which failure is preferable is a judgement call. Given that these stop mistakes rather than adversaries, the false positive is the better one to keep: a blocked comment is visible and one edit away, a silently permitted push is neither.

## `probe-hooks.sh`

Keep it. It found two defects while being written, and it is the only place the two cases above can be recorded so they do not return. Dropping it means the fixes for them ship untested.

bgunyel:
Both defects were real. Reproduced against the committed scripts before changing anything, including the part that stings most — `no-commit-to-main.sh`, with the anchor I called the looser one, blocks the indented push to main that `no-git-push.sh` waves through:

```
ALLOW  if/then + indented git push origin dev-05   (no-git-push.sh)
BLOCK  if/then + indented git push origin main     (no-commit-to-main.sh)
```

Fixed in 74c560c.

**Anchor.** Now `(^[[:space:]]*|[;&|(][[:space:]]*)`. You were right that this is not a free change, and right about which failure to prefer. The trade is now written down in both files and asserted in the suite, rather than being made silently and described as an improvement:

```
BLOCK  multi-line -b string continuing with a push    [accepted false positive]
BLOCK  multi-line -b string continuing with a merge   [accepted false positive]
```

Those two are probes with expected `BLOCK`, so a later change that makes them pass has to be a decision someone takes knowingly.

**Wrapper rule.** Added to `no-pr-decisions.sh`. Two shapes re-admit the raw command: a shell wrapper, and `gh api` reading a heredoc — the graphql case you identified, where the ordinary way to send a mutation was also the way to hide one.

**A third defect, mine, found while fixing yours.** I first tried to unify the two files by re-admitting the raw text instead of keeping a separate rule. That regressed three cases the old dedicated rule had caught:

```
FAIL want=BLOCK got=ALLOW  bash -c "git push origin dev-05"
FAIL want=BLOCK got=ALLOW  sh -c "git push"
FAIL want=BLOCK got=ALLOW  eval "git push"
```

Re-admission is sufficient for a heredoc, whose payload sits at the start of a line, and useless for `sh -c "..."`, whose payload sits inside quotes where no command position exists. Both files now also match *without* the anchor, and only once a wrapper has been found. The suite caught this immediately, which is the argument for it better than anything I wrote in the PR description.

**76 probes**, up from 63, all passing. Seven new regressions for the two defects above, two for the accepted false positive.

`probe-hooks.sh` stays.

One open question, unrelated to your review and worth settling before this merges: once these are live in the main checkout I can no longer push any branch, so I can no longer open a PR either — `gh pr create` needs the branch on the remote first. Today's flow becomes: I commit, you push, I open the PR. If you would rather keep agent-opened PRs working, the narrow change is to permit pushing a branch that is neither `main` nor a `dev-NN`, and refuse everything else.

bgunyel:
Added in a5e8fd1 — an agent may push the branch of the linked worktree it is working in, and nothing else. This restores the flow the blanket rule had ended: `gh pr create` needs the branch on the remote first.

**Keyed on where the command runs, not what the branch is called.** Worktrees are made two ways here — `EnterWorktree` prefixes the branch `worktree-`, `git worktree add -b` does not, and this branch has no prefix — so a name rule would be inconsistent between them, and a name can be chosen to match anyway.

```
--git-dir         .git/worktrees/hooks-guard     (linked worktree)
--git-common-dir  .git
```

Equal in the main checkout, different in a worktree.

**Five conditions, each refusing with its own reason:** runs in a linked worktree; that worktree is not on `main` or a `dev-NN`; names no other local branch, as a plain token or as `refs/heads/<name>`; does not `cd` first; does not redirect git with `-C` / `--git-dir` / `--work-tree`.

The last two close the blind spot `no-commit-to-main.sh` still has — the hook is judged on the *session's* directory, not the directory the command will run in. Tolerable while every push was refused; not once one is permitted.

A push inside `sh -c`, `bash -c`, `eval` or a heredoc fed to a shell stays refused with no exception: the payload sits in quotes where no command position exists, so the destination cannot be read, and permitting a push whose target is unknown is not the same as permitting this branch.

**One defect the suite caught while writing this.** Tokenising on a list of separators left a closing quote attached, so `-b "...main"` produced the token `main"`, missed the exact match, and a refused destination read as permitted. Splitting is now on everything a branch name *cannot* contain.

**76 probes.** The own-branch expectation is read from the environment rather than written as a literal, because whether that push is permitted is precisely what the hook decides — from a linked worktree the seven own-branch probes expect ALLOW, from the main checkout the identical commands expect BLOCK, and the suite prints which context it ran in:

```
context: linked worktree on hooks-push-and-pr-guards -- a push of this branch is permitted
```

Twelve further probes pin what the exception does not reach: `origin main`, `origin dev-05`, `HEAD:main`, a forced push to `dev-05`, a `cd` first, `-C` and `--git-dir` redirects, and all four wrapper forms.

**Worth a decision before merge:** force-pushing this worktree's own branch is currently permitted, which rewrites the history an open PR is showing. I left it that way because the ask was to permit pushing this branch — refusing `--force` and `--force-with-lease` is a one-line change if you'd rather.

**What I could not verify from here:** the main-checkout refusal path. This session is confined to the worktree, so I could not run the hook with the main checkout as its directory. The allow path is proven directly above; the refuse path rests on git's invariant that the two dirs are equal outside a linked worktree. Running `bash .claude/hooks/probe-hooks.sh` from the main checkout after merge confirms it — the banner will say so and the seven own-branch probes will flip to expecting BLOCK.

bgunyel:
Re-tested against `74c560c` + `a5e8fd1`. Both reported defects are fixed:

```
BLOCK  indented git push origin dev-05        BLOCK  bash -c 'gh pr merge 35'
BLOCK  indented gh pr merge 35                BLOCK  sh -c / eval  gh pr merge 35
BLOCK  bash -c 'git push origin main'         BLOCK  gh api graphql <<EOF mergePullRequest EOF
```

**The main-checkout path is verified.** Running the hook with the main checkout as its directory (on `dev-05`): `git push -u origin dev-05`, bare `git push`, `git push origin worktree-issue-27-three-shape-pool` and `git push --all origin` all refuse. The invariant that `--git-dir` equals `--git-common-dir` outside a linked worktree holds in practice, not just in principle.

## The exception is an allowlist described, a denylist implemented

The rule reads *"an agent may push the branch of the linked worktree it is working in, and nothing else."* The final loop walks local branches and refuses any that appear as a token — so a push that names **no** branch is invisible to it. From the worktree on `worktree-issue-27-three-shape-pool`:

```
ALLOW  git push --all origin                        advances remote dev-05 from this worktree's local ref
ALLOW  git push --mirror origin                     deletes every remote branch absent locally, closing open PRs
ALLOW  git push origin +refs/heads/*:refs/heads/*   wildcard, forced
ALLOW  git push origin --tags                       tags are not branches, so the loop never looks
ALLOW  GIT_DIR=/other/.git git push origin main     the env form of the redirect that -C / --git-dir close
ALLOW  pushd /some/repo && git push                 the cd guard does not cover pushd
```

`--all` alone defeats the rule: one command from any worktree pushes `main` and `dev-05`. `--mirror` is worse, because it deletes rather than advances.

## Why nothing behind it catches these

Two backstops might be assumed, and neither applies to `dev-NN`.

**The ruleset is not in force.** `GET /repos/bgunyel/clause-and-effect/branches/main/protection` returns `404 Branch not protected`; the `main-branch-protection` ruleset exists with `deletion`, `non_fast_forward` and `pull_request` rules, and sits at `enforcement: "disabled"`. Rulesets created through the UI start disabled unless set Active.

**And even enabled it covers `main` only** — its condition is `include: ["~DEFAULT_BRANCH"]`.

Extending it to `dev-*` does not work either, for a reason worth recording: **an agent pushes as `bgunyel`.** A server-side rule cannot distinguish Bertan from an agent running under his credentials, so a ruleset on `dev-*` either blocks both or allows both, and naming him a bypass actor hands the agent the same bypass. The policy "Bertan develops freely in dev branches, agents may only open PRs to them" has no server-side expression.

The linked-worktree check is the only signal anywhere that tells the two apart. That makes this hook the sole enforcement for `dev-NN`, not a convenience layer in front of a server rule.

## Suggested change

Invert the final check. The push must positively name the current branch or `HEAD`; refuse `--all`, `--mirror`, `--tags`, `--follow-tags`, and any refspec containing `*`. An allowlist cannot be walked past by a spelling that names nothing.

`GIT_DIR=` as an environment prefix belongs with the `-C` / `--git-dir` / `--work-tree` refusals it currently sits outside of, and `pushd` with `cd`.

For `main`, the ruleset is the right mechanism and only needs enabling — there the policy is identical for Bertan and agents, so a rule that blocks both is exactly correct.

bgunyel:
All seven reproduced against `a5e8fd1` before I changed anything. Fixed in 8b3e332.

You were right about the shape of the mistake: described as an allowlist, implemented as a denylist. The check walked local branches and refused any that appeared as a token, so naming nothing was the one thing it could not see.

**Inverted.** A push must positively name this branch — first bare token after `push` is the remote, every later one a refspec, and both sides of each must be this branch, `refs/heads/<this branch>`, or `HEAD`. Refused by name: `--all`, `--mirror`, `--prune`, `--tags`, `--follow-tags`, any refspec holding a wildcard, and deletion whether spelled `--delete` or as an empty source. With no refspec the destination comes from `push.default`, so `matching` — the one value that would carry other branches along — is refused too.

**`GIT_DIR=` was worse than unrefused: it was undetected.** An environment assignment precedes the command, so `git` was not at a command position and the push never reached the redirect rule at all. That is why the fix is in the anchor rather than in the rule. Every anchor in *both* files now admits a `VAR=` prefix, which closed the same evasion on the decisions side — `FOO=1 gh pr merge 35` was passing, and is now a probe. `pushd` joins `cd`.

**93 probes**, up from 76. Eight for the forms naming no branch, four for the redirect and `cd` shapes, and five asserting an ordinary push of this branch is still admitted — bare, with a remote, with `HEAD`, with an explicit refspec, and with a push option taking a separate value — so the inversion is shown not to have over-tightened.

Your point about `dev-*` is recorded in the file header, because it is the thing a later reader is most likely to get wrong:

> an agent pushes as `bgunyel`, so no server-side ruleset can separate the two. A rule on `dev-*` blocks both or allows both, and naming Bertan a bypass actor hands the agent the bypass.

That reframes this file from a convenience layer into the enforcement for `dev-NN`, and it is written down as such.

**On the ruleset — its state has changed since you wrote.** It now reads:

```
id 22380642  main-branch-protection  enforcement: active
conditions.ref_name.include: ["~DEFAULT_BRANCH"]
rules: deletion, non_fast_forward, pull_request
```

So the enabling is done. `/branches/main/protection` still returns 404, but that endpoint reports *classic* branch protection, which is a separate mechanism from rulesets — a 404 there is expected even while the ruleset is active, so it is not a reliable check for this. The `~DEFAULT_BRANCH`-only condition is exactly as you described, which is what makes `main` the half that server-side rules can carry and `dev-NN` the half they cannot.

I have not touched repository settings, and would not without asking.

**Still open, unchanged from my last comment:** force-pushing this worktree's own branch is permitted, which rewrites the history an open PR is displaying. `--force` and `--force-with-lease` join the refused list in one line if you want them there.

bgunyel:
Forced pushes now refused, in d2efb05. This closes the item left open on the last two comments.

Refused in every spelling: `--force`, `-f`, the short form bundled with other single-letter options as `-fu`, `--force-with-lease` with or without a value, `--force-if-includes`, and a leading `+` on the refspec — the same instruction written as punctuation, which the refspec loop had been quietly stripping before validating what was left.

`--force-with-lease` is refused with the rest rather than kept as the safe form. The lease guards against clobbering work someone else pushed in the meantime; it does not guard against rewriting a branch under the person reviewing it. On a solo repository with an agent pushing as the owner, the first hazard barely arises and the second is the whole point.

**One defect this surfaced.** The option checks were matching anywhere on the line rather than within the push's own arguments. Refusing `-f` that way would have made `rm -f notes.md && git push` read as a forced push — and `ls --all && git push` was already reading as `--all` before this commit, so the same flaw was live in 8b3e332. Every option check now reads only the arguments after `push`, up to the next shell separator. Pinned:

```
ok  ALLOW  rm -f before an ordinary push
```

**101 probes**, up from 93. The eight new ones cover each force spelling; the ninth is the false positive above.

Nothing is open from my side now. The exception is: from a linked worktree, on a branch that is neither `main` nor `dev-NN`, a non-forced push that positively names that branch — and nothing else.

bgunyel:
Re-tested against `8b3e332` + `d2efb05`. Everything claimed is delivered — all seven bypasses closed, and the ordinary shapes still admitted, including the two false positives `d2efb05` fixed:

```
BLOCK  --all  --mirror  --tags  wildcard refspec  GIT_DIR=  pushd  FOO=1 gh pr merge 35
BLOCK  --force  -f  -fu  --force-with-lease[=value]  --force-if-includes  +refspec  --force after the refspec
BLOCK  origin main   HEAD:main   HEAD:refs/heads/main   :main   --delete origin main
ALLOW  git push · git push -u origin <branch> · git push origin HEAD
ALLOW  rm -f notes.md && git push · ls --all && git push
```

## The argument scoping in d2efb05 opens a new hole

A backslash-newline continuation after `push` defeats every option and destination check added in `8b3e332` and `d2efb05`:

```
ALLOW  git push \
         --mirror origin

ALLOW  git push \
         --all origin

ALLOW  git push \
         --force-with-lease origin worktree-issue-27-three-shape-pool

ALLOW  git push \
         origin main
```

The mechanism: the scope runs from `push` to the next shell separator, a newline ends it, and **an empty scope falls through to the bare-`git push` case, which is the permitted one.** A trailing backslash therefore makes any push look bare.

The branch denylist still spans lines — `git push origin \` newline `main` is refused — so what the continuation defeats is precisely the allowlist inversion, not what preceded it.

`--mirror` is the one to weigh: reached this way it deletes every remote branch absent locally, which closes open PRs, and it is one wrapped line away.

This is the same defect class `d2efb05` itself fixed, mirrored. That commit found the option checks matching too *wide* — `ls --all && git push` reading as `--all` — and narrowed them to the push's own arguments. The narrowing now cuts at a newline that is not a separator at all. `no-pr-decisions.sh` is unaffected: `gh pr review 35 \` newline `--approve` is refused.

**Suggested fix.** A backslash-newline is a line continuation, not a command separator: join continued lines before scoping. And an empty argument scope should not fall through to the bare-push branch — after a trailing backslash it means "the arguments are elsewhere", not "there are none".

## Minor: the remote is never checked

```
ALLOW  git push git@github.com:someone/else.git HEAD
```

The refspec is legitimately this worktree's branch, so the allowlist admits it; nothing requires the remote to be `origin`. This reaches none of Bertan's branches, so it is a note rather than a defect — but a typo'd or invented remote is the shape of mistake this file exists for, and pinning the remote token to a name from `git remote` is one line.

bgunyel:
Both reproduced against `d2efb05`, both fixed in d4acf8d.

Your reading of the mechanism was exact, including the part that makes it dangerous rather than merely wrong: **an empty scope is the bare-push case, which is the permitted one.** So the failure mode is not "the check misses something", it is "the check reports the safest possible answer". That is the worst direction for it to fail in, and it is why a wrapped `--mirror` was one line away.

Continuations are now joined before anything is matched. In **both** files — the sibling's checks do span lines and were unaffected, exactly as you said, but the one previous time the two diverged it was because a rule lived in only one of them, so they are kept in step on purpose rather than on evidence.

You also named the class correctly: this is `d2efb05` mirrored. That commit found the option checks matching too *wide* and narrowed them to the push's own arguments; the narrowing then cut at a newline, which is not a separator when a backslash precedes it. Fixing one direction opened the other.

**A trailing backslash with nothing after it stays ALLOW**, and I think that is right rather than a residue: it is a continuation of nothing, so the command is a bare `git push` of this branch. It is now a probe with that reasoning attached, so it reads as a decision.

**One regression the suite caught immediately.** Checking the remote at all revealed that the scope did not end at a closing paren, so the `)` of `(git push)` was read as the remote name and an ordinary subshell push started failing. The scope now ends there too.

**The remote is checked.** Both `git push git@github.com:someone/else.git HEAD` and `git push upstream HEAD` are refused; the first bare token must name a remote this repository has. You filed it as a note rather than a defect and that was the right weight — it reaches none of your branches — but an invented remote is the shape of mistake this file exists for, so it is in.

**111 probes**, up from 101. Five for the continuation forms, two asserting a continued *legitimate* push still works, three for the remote, and the `(git push)` regression.

bgunyel:
Re-tested against `d4acf8d`. Both previous findings are fixed and nothing over-tightened:

```
BLOCK  continuation --mirror / --all / origin main   arbitrary remote URL   unknown remote name
ALLOW  (git push)   continued legitimate push   bare push   quoted branch   HEAD:refs/heads/<branch>   --set-upstream
```

I also checked the interaction I expected to break — a trailing backslash on the line before a heredoc terminator, which could have joined the terminator into the body and hidden a later command. It refuses in both shapes tested, so joining before dropping is in the right order.

## Only the first push on a line is validated

```
ALLOW  git push origin worktree-issue-27-three-shape-pool; git push origin main
ALLOW  git push origin worktree-issue-27-three-shape-pool && git push --all origin
ALLOW  git push && git push --force origin worktree-issue-27-three-shape-pool

BLOCK  git push origin worktree-issue-27-three-shape-pool
       git push --mirror origin                                  (newline, so a second line, so checked)
```

The scoping finds the first `push`, validates its arguments up to the next separator, and stops; it never looks for a second occurrence. A newline still works because the scan is line-based, but `;` and `&&` are the two commonest ways a command gets chained, and `git push origin my-branch && git push --tags` is an ordinary thing to write.

## Two smaller ones

```
ALLOW  echo `git push --mirror origin`         backticks; the $( ) spelling is caught
ALLOW  echo origin | xargs git push --mirror   exotic — a note, not a defect
```

The backtick case is the same asymmetry that `GIT_DIR=` had against `-C`: one spelling of a construct closed, its twin left open.

## The pattern is now four rounds old

Every defect found here, by either of us, has been the same defect: **how far around a matched token to look.**

- column-zero anchoring missed indented commands
- wrapper re-admission was sufficient for heredocs and useless for `sh -c`
- the option scope matched the whole line, so `ls --all && git push` read as `--all`
- narrowing it cut at a newline, so a continuation made every push look bare
- and it covers one occurrence, not all of them

Two of those were introduced by the fix to the previous two. The suite is at 111 probes and catches regressions well, but a probe suite only tests the shapes someone thought of, and every miss has been a shape of **shell syntax** rather than a shape of git.

That argues for one shared tokeniser rather than another regex: a single function both files call, answering where a command starts, where its arguments end, what a line continuation is, and **how many commands the string contains** — with the probes pointed at that function directly. Right now the rule is re-derived in a dozen expressions across two files, which is why fixing one keeps opening another.

bgunyel:
Taken, in full. 3d61363 adds `lib/command-scan.sh` and rewrites both hooks on top of it.

Your diagnosis was the useful part, more than either individual finding. Five defects, one question — how far around a matched token to look — answered differently in about a dozen expressions across two files, which is why fixing it in one place kept opening it in another. And every miss a shape of *shell syntax*, none a shape of git.

**Three functions, and the hooks now match at `^`:**

- `cs_normalise` — heredoc bodies dropped, continuations joined, in that order
- `cs_split` — one command per line, everything before the command word removed
- `cs_git_args` — a git subcommand's arguments, with **exit status** saying whether the command was one, so `git push` (empty arguments, the permitted shape) is distinguishable from "not a push" without a caller re-deriving the rule

Neither hook describes a command position anywhere any more.

**The three open items closed as a consequence rather than one at a time**, which is the part that makes the refactor worth its size:

- *Only the first push validated* — the loop walks commands instead of scanning for one, so `git push origin <branch> && git push --all origin` is refused on its second half.
- *Backticks* — a separator in the tokeniser, so both spellings close together. It was open in **both** files, incidentally: `` echo `gh pr merge 35` `` was allowed too, which your note did not cover.
- *xargs* — wrapper words and their options are stripped beside environment assignments. One line there; it would have been two more regexes in each hook.

**The split is deliberately more eager than a shell's.** Over-splitting yields extra command candidates, which can only refuse more — it never hides one. For a guard whose failure mode has twice been to report the *permitted* answer, that is the safe direction to be wrong in, and it is written down as such.

**One regression, caught by the suite in the same run that introduced it.** I put the new "is there a push at all" gate before the wrapper check; a wrapped push has no command word to find, so the gate answered no and all four wrapper forms went through. Reordered.

**134 probes**, up from 111. Twelve aim at the tokeniser directly rather than through a hook — splits, prefix stripping, continuation joining, heredoc dropping, and both exit statuses of `cs_git_args` — with the expected output written as literals, which is what you asked for when you said the probes should point at the function.

bgunyel:
Verified against `3d61363`. The refactor holds.

```
Round-4 findings, all closed:
  BLOCK  git push origin <branch>; git push origin main
  BLOCK  git push origin <branch> && git push --all origin
  BLOCK  git push && git push --force origin <branch>
  BLOCK  echo `git push --mirror origin`
  BLOCK  echo `gh pr merge 35`
  BLOCK  echo origin | xargs git push --mirror

Regression sweep, 19 prior BLOCK cases:  all still refuse
  --all · --mirror · --tags · wildcard refspec · GIT_DIR= · pushd · git -c
  --force · -fu · --force-with-lease · origin main · HEAD:main · :main
  --delete origin main · unknown remote · remote URL · continuation --mirror
  indented push · bash -c

Over-tightening check, 16 ALLOW cases:   all still permit
  bare · -u origin <branch> · origin HEAD · quoted branch · HEAD:refs/heads/<branch>
  --set-upstream · -o ci.skip · (git push) · continued legitimate push
  rm -f && push · ls --all && push · git status && push
  gh pr create · gh pr review --comment · gh issue close · heredoc prose

probe-hooks.sh, run from a linked worktree:  ALL PROBES PASSED
```

Over-tightening was the real exposure in this commit. A split deliberately more eager than a shell's buys safety by refusing more, and an agent that cannot push its own branch will work around the guard rather than within it. Nothing in the legitimate set moved.

**The three findings closed as a consequence of the abstraction rather than one at a time**, which is the test of whether it was the right abstraction. And the backtick hole was open in `no-pr-decisions.sh` too — `` echo `gh pr merge 35` `` — which my note did not cover and this found on its own.

From my side this is done, and I would merge it.

## One thing left, and it is not code

The guarantee now exists only as 743 lines of shell and 134 probes. Nobody can answer *"what can an unattended agent do to this repository?"* without reading them — including whoever reads this in three months.

If AFK sessions are the working mode this file exists to enable, that boundary belongs in `CLAUDE.md` as a paragraph a human can audit in thirty seconds:

> An agent may push the branch of the linked worktree it is working in — non-forced, positively named, to a remote this repository has — and nothing else. It may open, comment on, edit and view pull requests. It may not merge one, review one with a verdict, close or reopen one, or create a release. `main` and `dev-NN` are Bertan's; `main` is additionally protected server-side by a ruleset requiring a pull request.

Plus a line naming whatever is deliberately left open, with the standard that excludes it. The code implements the boundary; nothing states it.

bgunyel:
Added in 55be4f0 — a `## What an unattended agent may do to this repository` section in `CLAUDE.md`, sited after Working conventions.

You were right that this is not code and right that it comes last. The hooks implement a boundary; nothing stated one, and 733 lines of shell is not an answer to a question a person should be able to ask in thirty seconds.

I kept your paragraph almost verbatim, and added two things the code can only imply:

**Why `dev-NN` cannot be carried anywhere else.** An agent pushes as `bgunyel`, so a rule on `dev-*` blocks Bertan too and a bypass actor hands the agent the bypass — the linked-worktree check is the only signal separating them, and `main` is carried by a ruleset precisely because there the policy is identical for both. That was your analysis two rounds ago and it is the fact most likely to be lost, because it explains an absence.

**That hooks see only the Bash tool**, so your own terminal is outside all of it. Without that line, refusing every push looks costly rather than free, and someone would eventually loosen it for the wrong reason.

**And the standard, not just the exclusion.** What is left open is named with the reason it is left open — these stop mistakes and not adversaries; a wrapped push is refused rather than assessed because a destination inside quotes cannot be read; prose whose continuation line begins with one of these commands is refused although it is only prose. Silence there would read as oversight, which is what the last five rounds were about.

No code changed. 134 probes still pass.

That closes everything raised on this PR. Eight commits, and it is yours to merge — which the hooks now guarantee I cannot do.

bgunyel:
## Second review: five defects, all fixed in `02a14d8`

A Codex review was run against this branch (`--base dev-05`) after the first
round of review comments had been settled. It found five defects. The assistant
reproduced each one against the hooks as they stood before fixing it, so every
item below is an observed exit status, not a reading of the code.

Four were bypasses of the boundary this branch introduces. One was the opposite
error — the guard refusing something `CLAUDE.md` explicitly allows.

### What was wrong

| # | Defect | Before | Where |
|---|---|---|---|
| 1 | A command after a control word is at a command position, and was not treated as one | `if true; then git push --mirror origin; fi` → **allowed** | `lib/command-scan.sh`, `cs_split` |
| 2 | `<<<` is a here-string, read as a heredoc whose terminator never arrives | `cat <<< "hello"` + `gh pr merge 35` → **allowed** | `lib/command-scan.sh`, `cs_normalise` |
| 3 | `<<-` strips leading tabs from its terminator; an exact comparison never matched one | `cat <<-EOF` / tab-indented `EOF` + `gh pr merge 35` → **allowed** | `lib/command-scan.sh`, `cs_normalise` |
| 4 | A bare push is answered by configuration, and the check read one of the four places it can be written | `git -c push.default=matching push` → **allowed** | `no-git-push.sh`, `check_push` |
| 5 | The merge and review endpoints were refused whatever the HTTP method | `gh api .../pulls/35/reviews` (a GET) → **refused** | `no-pr-decisions.sh` |

The first three were the same question this branch already claims to answer
once — how far around a matched token to look — so all three were fixed in
`lib/command-scan.sh` rather than in the hooks. The reported class for #1 was
also narrower than the real one: `then`, `do`, `else`, `elif`, `{` and `!` all
defeated the `^` anchor. Only `case` survived, incidentally, because `)` is
already a separator.

### Why #4 was not fixed by reading configuration more carefully

A push naming no refspec is sent where `push.default`, a `remote.<name>.push`
refspec, or the branch's upstream says — and `-c` sets any of them for one
command, past whatever the hook reads back afterwards. Reading all four would
still not have closed it: an agent may run `git config`, and nothing here
refuses it, so a rule resting on a configured value can be arranged around one
command earlier.

So the destination is required in the command instead, which is what
`CLAUDE.md` already asked for — a push "positively naming that branch".

### Three trades, taken knowingly

Recorded here, in the commit message, in the comments beside the code, and as
probes, so that a later change that reverses one is a decision rather than a
regression:

1. **`git push` and `git push origin` were permitted and are now refused.**
   Write `git push origin <branch>`. Two probes that asserted the old behaviour
   now say `BLOCK`.
2. **A quoted string holding a separator and then a control word reads as a
   command**, so `git commit -m "wait; then git push --all origin"` is refused.
   The price of #1's fix; it sits beside the older false positive of the same
   kind, in the `ACCEPTED false positive` section.
3. **A `gh api` read wrapped in a shell is still refused.** Inside
   `bash -c '...'` the method cannot be read any more than the endpoint can, so
   the wrapper rule stays blunt. Run it unwrapped.

### On the probe suite

134 probes → 168, all passing from the linked worktree.

Worth stating plainly rather than only counting: **the suite was green while
`if true; then git push --mirror origin; fi` was permitted.** A probe suite is
evidence about the cases it names and about nothing else, and the suite's own
header now says so. All four bypasses were found by an outside reviewer asking
questions the suite did not ask.

### Still outstanding

- `bash .claude/hooks/probe-hooks.sh` has still never been run **from the main
  checkout**. It was verified from the `hooks-guard` worktree, where the seven
  own-branch probes read `ALLOW`; from the main checkout the identical commands
  should read `BLOCK`, and the suite prints which context it ran in. That
  assertion rests on git's documented invariant rather than on an observed run.
- The hooks remain inert until this PR merges into `dev-05`.
- No `docs/dev-log/` entry was added, for either session on this branch.

https://claude.ai/code/session_01RaFbXHgAAHKBWsynwqCEU6

bgunyel:
Ran both hooks from `02a14d8` against `stdin` shaped like the real payload, from a linked worktree on `feature-x` and from the main checkout. `probe-hooks.sh` is green. Four ways past the boundary survive it, and one of them is the same defect the header of `command-scan.sh` says was hunted twice.

## 1. A `<<` inside a quoted string is read as a heredoc, and everything after it is dropped

`cs_normalise` finds a heredoc opener anywhere on the line, quotes included. When the "terminator" it takes never arrives, `ind` stays set to EOF and **every remaining line is dropped** — so both hooks go blind for the rest of the command.

```
ALLOW   git commit -m "hooks: fix <<EOF handling in cs_normalise"\n    git push --all origin
ALLOW   gh pr comment 35 -b "the << operator confused it"\n    gh pr merge 35
ALLOW   git commit -m "left shift << done"\n    git push --all origin
ALLOW   gh issue comment 1 -b "see << notes"\n    git push --mirror origin
```

What `cs_normalise` emits for the first one is one line:

```
git commit -m "hooks: fix <<EOF handling in cs_normalise"
```

The push is gone. Bash runs it — `<<` inside double quotes is text, not a redirection — so this is a divergence from the shell, not a shared limitation.

This is the third instance of the same question. The file already records `<<<` and `<<-` as answered wrongly, and says why it matters: *"Dropping is the one step here that hides commands rather than exposing them, so what counts as a heredoc has to be exact in both directions."* Getting it exact in both directions is the hard version. The cheap version is to make the step fail safe, which matches the direction the file takes everywhere else — **if the terminator never arrives, the drop was wrong; re-emit what was held.**

```awk
    ind {
      line = $0
      if (dash) sub(/^\t+/, "", line)
      held[++nheld] = $0
      if (line == d) { ind = 0; nheld = 0 }
      next
    }
    ...
    END { for (i = 1; i <= nheld; i++) print held[i] }
```

Tested: all four rows above flip to BLOCK, the heredoc commit message naming `git push origin main` and `gh pr merge 35` still ALLOWs, and `probe-hooks.sh` stays green. The cost is a genuinely unterminated heredoc now being scanned — a command bash would refuse to run anyway, refused in the direction this file has chosen throughout.

## 2. `gh pr review -ab "lgtm" 35` approves, and is allowed

```
BLOCK   gh pr review -a 35
BLOCK   gh pr review --approve 35
ALLOW   gh pr review -ab "lgtm" 35
ALLOW   gh pr review 35 -ab lgtm
ALLOW   gh pr review -rb "no" 35
```

`gh` bundles shorthand flags. Confirmed against the real binary — `gh pr view 35 -cq .number` binds both `-c` and `-q` and fails only on `--jq` needing `--json`, so the bundle was parsed. `-a` and `-r` are `--approve` and `--request-changes`.

`no-git-push.sh` already got this right for the same problem: `-[A-Za-z]*f[A-Za-z]*` catches `-fu`. The sibling didn't. That is the asymmetry the description tells about the wrapper rule, in a second place.

## 3. A flag before the subcommand hides the subcommand

```
ALLOW   gh pr --repo o/r merge 35
```

Every `gh pr` rule requires the subcommand as the third word. Cobra takes it anywhere — verified with `gh pr --repo cli/cli list -L 1`, which lists. The rules need to skip flag tokens between `pr` and the subcommand.

## 4. Closing a PR and publishing a release are blocked in the CLI spelling and open through `gh api`

```
BLOCK   gh pr close 35
ALLOW   gh api -X PATCH repos/o/r/pulls/35 -f state=closed

BLOCK   gh release create v1
ALLOW   gh api -X POST repos/o/r/releases -f tag_name=v1

ALLOW   gh api graphql -f query="mutation{closePullRequest(...)}"
```

The API endpoint list is `/pulls/*/merge` and `/pulls/*/reviews` only. The write test itself works — all three are correctly identified as writes — but no rule matches them. *"The direct API path is matched too"* holds for merging and for reviewing; it does not hold for closing or for releasing, which the same file refuses one spelling earlier.

`/releases` and `closePullRequest` are free to add. `PATCH /pulls/N` is not: it is also how `gh pr edit` retitles, which is allowed. Either match the field (`state=closed`, `state=open`) or accept the gap — but then say so, because CLAUDE.md currently reads as though the boundary is spelling-independent.

## Minor

`for TOK in $ARGS` in `check_push` is unquoted, so the argument list is glob-expanded against the worktree. `*` is refused by name one line earlier, `?` and `[...]` are not. Both currently land on BLOCK, because an expansion that matches a file does not name the branch and one that matches nothing stays literal — so this is a latent surprise rather than a hole. `set -f` before the loop closes it.

`\gh pr merge 35` and `python -c "...os.system('gh pr merge 35')"` both pass. Those sit inside *"stops mistakes, not adversaries"* and I would leave them.

## `probe-hooks.sh`

Keep it, and this round is the argument for keeping it rather than against. The suite was green while all four findings above were live — which its own header already concedes about the previous round — and that is an argument for more probes in it, not for deleting the only place they can be written down. Findings 1 through 4 are each one line there.

## On the description

It is accurate about what it claims except for finding 4, where *"The direct API path is matched too"* covers less than the section implies. Worth narrowing to merge and review, which is what the code does.

bgunyel:
## Third review: four more, fixed in `b625d64`

All four reproduced against `02a14d8` before anything was changed, and the
`gh pr edit` control was checked alongside each fix. The probe suite was green
the whole time — the second round in a row that has been true, which is now in
the suite's own header rather than in a commit message that scrolls away.

### 1. A `<<` inside a quoted string — the one that matters

```
git commit -m "hooks: fix <<EOF handling in cs_normalise"
git push --all origin
```

`cs_normalise` emitted only the first line. The push was gone, and bash would
have run it — `<<` inside double quotes is text, so this was a divergence from
the shell rather than a limitation shared with it. Reachable by writing a
commit message about this file.

This was the **third** wrong answer to what counts as a heredoc. The first two
rounds answered it by making the match more exact — excluding `<<<`, tracking
`<<-`. Three silent failures, every one in the permitting direction, is
evidence about the question and not about the answers: looking more carefully
is exactly what the previous two attempts were.

So the drop is a fail-safe now, as proposed in the review:

```awk
    ind {
      line = $0
      if (dash) sub(/^\t+/, "", line)
      held[++nheld] = $0
      if (line == d) { ind = 0; nheld = 0 }
      next
    }
    END { for (i = 1; i <= nheld; i++) print held[i] }
```

A heredoc that *does* terminate still has its body dropped — the older
commit-message probes still `ALLOW`, and that is what says the fail-safe did
not simply disable the drop.

### 2, 3. Two more of the sibling asymmetry

- `gh pr review -ab "lgtm" 35` approved. gh bundles shorthand flags; the rule
  listed `-a` and `-r` as whole tokens. `no-git-push.sh` had already answered
  this for `-fu`. Only single-dash bundles are scanned for `a` or `r` — a long
  flag would match on any letter it contains, and `--repo` would read as
  `--request-changes`. Probed both ways.
- `gh pr --repo o/r merge 35` merged. Cobra resolves the subcommand at the
  first non-flag argument. `-R`/`--repo` consumes its value, which would
  otherwise be read as the subcommand and hide it just as well.

### 4. Close and release through `gh api`

`PATCH /pulls/N -f state=closed`, `POST /releases` and `closePullRequest` were
all correctly identified as writes and then matched by no rule. Fixed by
widening the code rather than narrowing the claim, so the description now
carries a table of what is matched instead of a principle.

`PATCH /pulls/N` is matched on the **field**, not the endpoint, because that is
also how `gh pr edit` retitles — asserted to stay allowed by a probe. That
assertion earned itself immediately: `gh pr edit` fails on this repository with
a Projects-classic GraphQL error, so this comment's own description update went
out as `gh api -X PATCH .../pulls/35 -F body=@file`, which the hook allows.

### Minor

`set -f` added to `no-git-push.sh`. Latent rather than a hole, as you read it,
and closed rather than left to become one.

`\gh pr merge 35` and the `os.system` route left alone, on your reading — both
sit inside *"this stops mistakes, not adversaries"*.

### Suite

168 probes → 207. Findings 1–4 each have a section, each with the `ALLOW`
controls that say what the fix did not break: retitling through the PATCH
endpoint, listing releases, a comment bundle carrying no verdict letter, and an
ordinary subcommand behind a flag.

Still outstanding, unchanged: the suite has never been run **from the main
checkout**, where the seven own-branch probes should read `BLOCK` rather than
`ALLOW`. The hooks stay inert until this merges into `dev-05`.

https://claude.ai/code/session_01RaFbXHgAAHKBWsynwqCEU6

bgunyel:
## Review: #37's criteria are met; three complete bypasses of `no-git-push.sh` are not

Every verdict below was produced by running the hook as a process — a
`tool_name`/`tool_input.command` payload on stdin, exit code 2 read as a
refusal — from the linked worktree `.claude/worktrees/hooks-guard`, so the
worktree exception is live. Same seam as the check suite. Nothing was pushed,
committed or posted anywhere else.

---

## Part 1 — Does this branch satisfy #37?

Yes. Seven of seven, six of them cleanly.

| # | Criterion | |
|---|---|---|
| 1 | `worktree.baseRef` set in the checked-in settings | ✅ `"worktree": {"baseRef": "head"}`; fork point verified in #37's closing comment (`68d489e` = `dev-05`, not `origin/main` `3c936c3`) |
| 2 | Both hook files carry the stopping rule | ✅ present in both, identical wording |
| 3 | CLAUDE.md drops "can rest on nothing else"; credentials stated as a choice | ✅ removed in `6416c8d`; replacement reads "a configuration choice and not a constraint" |
| 4 | "Deliberately left open" names the `.claude/` gap and the standard that excludes it | ✅ |
| 5 | CLAUDE.md's boundary section reduced to its operative paragraph | ⚠️ arguable — below |
| 6 | `docs/adr/0001` exists, principle + rejected alternative, no inventory | ✅ |
| 7 | The check suite passes, with the same count as before this ticket | ✅ **measured, not assumed** |
| 8 | Everything lands on this branch | ✅ |

Criterion 7 was re-measured rather than taken from the ticket. The suite at
`b625d64` — the count #36 quotes — was extracted and run beside the current one
in the same worktree:

```
checks at b625d64:  207 ok, 0 FAIL
checks at 2fad130:  207 ok, 0 FAIL
```

Unchanged, and the diff touches no Python, so `make test` is unchanged by
construction.

**Criterion 5 is the soft one.** The section runs to 38 lines: the operative
paragraph, a credentials paragraph, and a 15-line "deliberately left open"
block. Criteria 3 and 4 mandate that extra content, so 5 cannot be met
literally. But the last block still argues rather than states — "closing it
would make every future hook change a two-person procedure for no gain against
the threat actually named" is ADR reasoning sitting in the file an agent reads
each session. Substantially met, not cleanly.

### The one claim #37 exists to fix, and did not

#37: *"This ticket makes the branch's claims true before it lands."*

**This pull request's own description still says "63 probes." The suite has
207.** The body itemises 4 + 4 + 55 = 63, and that arithmetic has been stale
since before `b625d64` — the very commit #36 measured 207 at.

The count is the smaller half. The description describes a branch that no
longer exists:

- **The whole content of #37 is absent from it.** No mention of
  `docs/adr/0001`, the CLAUDE.md rewrite, the stopping rule, or
  `worktree.baseRef`. Seven files of work described in the language of three.
- "**Two rounds** of false positives" — there have been six-plus, and
  `command-scan.sh`'s own header says so at length.
- "**Reviewer's call:** this third file is the one thing here that was not asked
  for" — no longer true. The ADR and the CLAUDE.md section are the fourth and
  fifth things, and they *were* asked for.

#36's sharpest line was that "the code on #35 has been read closely six times,
and what it asserts about itself had not been." #37 fixed that for CLAUDE.md,
the hook headers and the ADR, and left it standing in the artefact a reviewer
reads first and the merge commit will carry.

Answer to the ticket: close #37. Rewrite this description before merging — it is
the ticket's own remit and a five-minute edit.

---

## Part 2 — Findings against the code

Out of scope for #37, and named here because they bear on whether this branch
should merge as it stands. All 207 checks pass; none of these is asked about.

### The root cause, first

Three of the five are one question answered in three places and correct in one:
**does this option take its value as a separate token?**

- `cs_git_args` handles `-c`/`-C` with a separate value — and not
  `--git-dir`/`--work-tree`/`--namespace`/`--exec-path`/`--config-env`, which
  it matches only in the `=` form.
- `cs_split`'s wrapper-word stripper handles an *attached* value (`-n1`) and not
  a separate one (`-n 1`).
- `no-pr-decisions.sh`'s `GHPR` handles `-R`/`--repo`/`--hostname` correctly.

Every failure is silent and in the permitting direction — the pattern
`command-scan.sh`'s header says the file exists to end. It stopped one step
short of ending it.

### 1. `git --git-dir <path> push` — the space form is invisible (blocker)

```
git --git-dir=/elsewhere/.git push --all origin    BLOCK   ← already checked
git --git-dir /elsewhere/.git push --all origin    ALLOW   ← same command
git --work-tree /elsewhere push --mirror origin    ALLOW
git --config-env push.default=PD push origin main  ALLOW
```

`cs_git_args` requires `=` on the long global options. The space form falls
through to the generic `-[^[:space:]]+` arm, which eats `--git-dir` and leaves
`/elsewhere/.git push …`, so `^push` never matches, `HAVE_PUSH` stays empty and
the hook exits 0 — **before** `--all`, `--mirror`, `--force` or the remote
check are reached. `git --git-dir <path>` is valid syntax (checked, git 2.43).

The tell that this is a defect and not a scoping decision: `check_push` already
greps for `(-C|--git-dir|--work-tree)([[:space:]]|=)` — it explicitly
anticipates the space form. That line is unreachable.

`no-commit-to-main.sh` does not catch it either; `git --git-dir /e/.git push
origin main` passes all six Bash hooks.

### 2. `timeout N …` bypasses both hooks (blocker)

```
timeout 60 git push --all origin    ALLOW
timeout 60 gh pr merge 35           ALLOW
```

`timeout` is not in `cs_split`'s wrapper list, and adding it is not sufficient:
its first argument is a duration, not a flag, and the stripper only consumes
leading `-options`.

### 3. `xargs -n 1` / `nice -n 10` — the value becomes the command word (blocker)

```
echo origin | xargs -n1  git push --mirror   BLOCK   ← already checked
echo origin | xargs -n 1 git push --mirror   ALLOW
nice -n10 git push --all origin              BLOCK
nice -n 10 git push --all origin             ALLOW
echo 35 | xargs -n 1 gh pr merge             ALLOW
```

`cs_split` emits `1 git push --mirror` and `10 git push --all origin`.

### 4. Bundled shell flags walk past the wrapper rule (major)

```
bash -c  "git push --all origin"   BLOCK   ← already checked
bash -lc "git push --all origin"   ALLOW
sh -ec   "git push --all origin"   ALLOW
bash -lc "gh pr merge 35"          ALLOW
```

Both wrapper regexes hardcode `(-c|<<)` as a standalone token. This is the third
appearance of the bundled-short-flag class this branch has already fixed twice —
`git push -fu`, then `gh pr review -ab`. Both fixes ruled that bundled shorthand
is a shape an agent writes without meaning to evade anything.

### 5. A wrapper anywhere refuses a permitted push (medium, false positive)

```
bash -c 'make test' && git push origin hooks-push-and-pr-guards   BLOCK
git push origin hooks-push-and-pr-guards                          ALLOW
bash -c 'make test' && gh pr list --state=open                    BLOCK
```

The wrapper branch requires a wrapper *somewhere* and a payload *somewhere*, not
the payload inside the wrapper. So an unrelated `bash -c` earlier on the line
refuses the one push the exception exists to permit, with a message — "git push
inside a shell wrapper" — describing something that is not happening. The check
`rm -f before an ordinary push` already asserts this principle for flags; it is
unapplied to wrappers.

### 6. Publishing a release is still spelling-dependent (medium)

```
gh api -X PATCH repos/o/r/releases/123 -f draft=false   BLOCK
gh release edit v1.0.0 --draft=false                    ALLOW   ← publishes the draft
gh release upload v1.0.0 asset.zip                      ALLOW   ← public assets
```

`(create|delete|delete-asset)` misses the two verbs that make an invisible
release visible. `02a14d8` fixed exactly this asymmetry for close/reopen; here
it survives inverted — the `gh api` path is covered and the convenient spelling
is not.

### Minor

- **The check suite is wired to nothing.** No `.github/workflows/`, no `make`
  target. The only artefact in this branch that produces evidence rather than
  argument runs when someone remembers to type it.

---

## Part 3 — What this means for #36's plan

These findings sit inside #36's Out of Scope, under:

> "Changing what the push hook refuses. **Its rules are settled** by six review
> rounds; stage 2 touches the pull-request hook only."

A seventh round found three complete bypasses of that hook, each permitting
`--all` and `--mirror`. "Settled" is a claim about itself of exactly the kind
#37 was created to correct, and it is the premise on which stage 2 declines to
touch the push hook.

The stopping rule admits all three on its own terms. `--git-dir=` and
`xargs -n1` are already checked as shapes an agent plausibly writes; the
space-separated spellings are the same commands typed the other way. They are
not constructions.

**Suggested disposition**

1. Close #37 — its criteria are met.
2. Rewrite this PR's description before merging.
3. Findings 1–3 are one function's worth of work: give `cs_split` and
   `cs_git_args` a shared option table that knows arity, instead of two
   hand-rolled regexes that each guess. That is the consolidation
   `command-scan.sh` was created for.
4. Either schedule that as a stage-0.5 ticket, or strike "its rules are settled"
   from #36's Out of Scope so stage 2 is not built on it.

Per the standard of care #36 proposes for stage 3: each of the six above should
arrive with a check that fails without its fix.

---

*Review produced by a Claude Code session at Bertan's request; every verdict
re-run against the hooks at `2fad130` rather than reasoned from the source.*
https://claude.ai/code/session_01XRX8DS8KXhZKppc86hFk97

PR #57, opened 2026-09-09
Refuse work on a worktree branch whose life is over, and arm the detectors that see it

### LINKED ISSUE #44: Refuse work on a worktree branch whose life is over, and arm the detectors that see it
## Parent

#36

## What to build

A worktree branch exists for one pull request. When that pull request merges,
the branch has served its purpose and should cease to exist — remotely by
`delete_branch_on_merge`, now enabled, and locally by Bertan's sweep. Nothing
stops work continuing on it in the meantime, and that failure has fired twice:
the `probe`→`check` rename was committed onto `hooks-push-and-pr-guards` after
PR #35 had already merged it, and `research/non-openrouter-response-bodies` sat
one commit ahead and nineteen behind with no pull request, loaded to revert
nineteen commits the moment anyone proposed it.

Refuse commands that add work to such a branch, keyed on running in a linked
worktree — the same keying the push hook uses, and deliberately not on the
branch's name.

**Two detectors, each covering the other's blind spot.**

*The upstream is gone.* With `delete_branch_on_merge` on, a merged worktree
branch is exactly a branch whose upstream has been pruned away. This reads the
remote's existence rather than the commits' ancestry, so no merge style defeats
it.

*`ahead == 0 && behind > 0`* against the active dev branch, as the fallback for
when nobody has pruned yet. This one rests on *merged implies ancestor*, which
is a property of merge style rather than of git — so `allow_squash_merge` and
`allow_rebase_merge` are disabled on the repository, making the assumption true
by configuration rather than by habit. An assumption a repository setting can
silently break is not an assumption; it is a bug with a delay.

**The active dev branch** is the highest-numbered `refs/remotes/origin/dev-*`.
No network: a hook has five seconds. The lexical sort is only correct because
`branch-hygiene` mandates two-digit zero-padding — name that dependency in the
header or use a version sort, so `dev-09`/`dev-10` does not become the next
defect. When no such ref exists — a fresh clone, or the rotation window after
the merged `dev-NN` is deleted and its successor is not yet pushed — the guard
**abstains** rather than refuses. These stop mistakes, not adversaries.

**The messages state what the hook can see.** `upstream: gone` fires only on the
genuinely merged case, so that message can say "merged" truthfully. The fallback
cannot distinguish a merged branch from a brand-new worktree branch that has not
committed while `dev-NN` moved on — the two are topologically identical, and
only a network call would separate them — so its message is about state: this
branch has no work of its own and the active dev branch has moved past it.

**Which commands.** Any command that adds a commit makes `ahead > 0` and retires
the fallback for that branch permanently, so the test is not "would an agent
plausibly write this" but "would this silently disable the rule". That is
commit, cherry-pick, revert, merge and `am`. `cherry-pick` is not marginal — it
is the command in this repository's own recovery procedure. Mid-operation
continuations such as `rebase --continue` are excluded, and the reason belongs
in the header: a refusal mid-rebase strands state the agent cannot exit, and a
guard that leaves the repository in a condition only a human can clear is worse
than the mistake it prevents.

**The one carve-out.** Under the fallback, `ahead == 0` means the branch is a
strict ancestor of the active dev branch, so merging or rebasing onto that
branch is a fast-forward: it creates no commit and masks nothing. Merge was in
the set for a hazard unreachable in that exact state, and refusing it there
deadlocks the branch — no commit, no catch-up, and removing a worktree is a
reserved act, so a timing race becomes a hand-off to a human. So under the
fallback, permit a merge or rebase **naming the active dev branch**, and refuse
one naming anything else, which would create a real commit and would mask.
Under `upstream: gone`, refuse both: merging into a branch that no longer exists
on the remote is meaningless.

**Arming.** Both detectors read remote-tracking refs and are exactly as fresh as
the last fetch — the fallback no less than the gone test, since `origin/dev-NN`
is a remote-tracking ref too. A hook cannot fetch. So a `SessionStart` hook runs
`.claude/hooks/report-stale-branches.sh`, whose first act is a pruning fetch:
the read-only half of the sweep is what arms the enforcing half. It prunes
explicitly in the script rather than relying on `fetch.prune`, because a guard
whose soundness depends on per-clone configuration is the push hook's own
objection pointed inward — and per-clone configuration is missing on exactly the
clone where it matters.

The script is read-only by name and by content. Removing a worktree or a branch
is Bertan's, per #41; the agent's half is to fetch, report and stop. It is named
`report-stale-branches.sh` and never `sweep-`, because the filename is where
that constraint survives the next reader.

## Acceptance criteria

- [ ] A commit on a worktree branch whose upstream is gone is refused, with a
      message naming the merge
- [ ] A commit on a worktree branch at `ahead == 0, behind > 0` is refused, with
      a message about state and not about merging
- [ ] `cherry-pick`, `revert`, `merge` and `am` are refused in the same states
- [ ] `rebase --continue` and other mid-operation continuations are permitted,
      with the reason written in the header
- [ ] Under the fallback, a merge or rebase naming the active dev branch is
      permitted; one naming any other branch is refused
- [ ] Under `upstream: gone`, merge and rebase are refused
- [ ] A fresh worktree branch at the active dev branch's tip is permitted
      (`ahead == 0, behind == 0`)
- [ ] A worktree branch carrying work of its own is permitted (`ahead > 0`)
- [ ] The main checkout is unaffected — the guard keys on the linked worktree
- [ ] The guard abstains when no `refs/remotes/origin/dev-*` exists
- [ ] `settings.json` carries a `SessionStart` entry running
      `report-stale-branches.sh`
- [ ] The script fetches with an explicit prune, reports stale worktrees and
      branches, and removes nothing
- [ ] The script fails open with a timeout: an offline session starts, and its
      report says the fetch failed
- [ ] Checks assert the arming properties as literals — that the script's fetch
      prunes, and that `settings.json` carries the entry
- [ ] `allow_squash_merge` and `allow_rebase_merge` are disabled on the
      repository
- [ ] Mutation-checked: deliberately breaking each rule turns its checks red
- [ ] The check suite passes

## Blocked by

Blocked by #43 — the guard reads repository state from the command's directory,
which the migrated file establishes and today's file does not.

## Notes

**Internal order.** Land `report-stale-branches.sh` and read its output once
before writing the guard's checks. #39 names the risk of a helper written before
its caller; this is that risk pointed the other way — a guard written against a
script nobody has run.

**Two soft edges, both deliberate, both belonging in the header.** The fetch
fails open, so an offline or slow session proceeds with neither detector armed;
that is announced by the report, which says the fetch failed. And permitting the
catch-up merge means a genuinely merged branch can be fast-forwarded to the tip,
after which both detectors read clean — a window lasting only a session in which
a merge lands after that session's fetch, closed at the next session start.

**The arming step is not self-announcing, and `.claude/` is unguarded.** #36
leaves that directory open on the grounds that an agent does not *mistakenly*
rewrite the hook that just refused it — breakage announces itself, because the
refusal stops coming. `report-stale-branches.sh` breaks that reasoning: it arms
enforcement rather than performing it, so dropping its prune changes nothing
visible. The exclusion is not reopened — extending the `Edit|Write` guard would
make every hook change a two-person procedure, which #36 declined for no gain
against the named threat. It is **re-argued**: anything in `.claude/` that arms
enforcement rather than performing it carries a check asserting its arming
property, which is the guard-code standard of #41 applied to the one file where
its usual justification fails.

That check announces at the next review rather than on the next push, because
**this repository has no CI** — there is no `.github/workflows/`, so the suite
runs only when someone runs it, and a broken `.claude/` could suppress the very
check that would report it. That is how all 207 existing checks already behave,
and it is far better than silence, but the ticket says which of the two it is
rather than leaving a reader to assume the stronger one.

Asserting `settings.json`'s content is a **second kind of check** in a suite
that today only drives hooks as processes and reads their exit codes. Small, but
it is a new seam and is named as one.

**Order.** This ticket and #43 are two small changes to one file, reviewable back
to back in one context, and #43 fixes a defect that has twice produced false
positives against Bertan's own work. That is why they come before #39 and #40 —
not the state of any hole, since both of the lifecycle hole's firings are
remediated in the same batch that files this.

bgunyel:
Closes #44.

A worktree branch exists for one pull request. When that pull request merges the
branch has served its purpose, and nothing stopped work continuing on it — a
failure that has fired twice here. This adds the guard that refuses it and the
SessionStart script that arms the guard.

## What is here

**`.claude/hooks/no-work-on-stale-branch.sh`** — two detectors, each covering
the other's blind spot. *The upstream is gone*, which with
`delete_branch_on_merge` is exactly what a merged worktree branch looks like,
and which reads the remote's existence rather than the commits' ancestry, so no
merge style defeats it. And *`ahead == 0 && behind > 0`* against the active dev
branch, as the fallback for when nobody has pruned yet. Keyed on running in a
linked worktree, the same keying `no-git-push.sh` uses and deliberately not on
the branch's name.

Refused: `commit`, `cherry-pick`, `revert`, `merge`, `am`, `rebase`.
Mid-operation continuations are permitted — a refusal mid-rebase strands state
the agent cannot exit. Under the fallback a merge or rebase **naming the active
dev branch** is permitted, because at `ahead == 0` that is a fast-forward: it
creates no commit and masks nothing, and refusing it would deadlock the branch.
One naming anything else is refused, and under `upstream: gone` both are.

**`.claude/hooks/report-stale-branches.sh`** — the read-only half of the
branch-hygiene sweep, run at `SessionStart`. Both detectors read
remote-tracking refs and are only as fresh as the last fetch, and a hook cannot
fetch, so this does it with an explicit `--prune` rather than trusting
`fetch.prune`. It fails open with a timeout and says so when it does. Run once
before its checks were written — per the ticket's internal order — it found two
merged worktrees still on disk.

**`check-hooks.sh`** — 347 checks, up from 207. Includes a second kind of check
the ticket names: assertions about a *file* rather than a process, because
`report-stale-branches.sh` arms enforcement rather than performing it, so
dropping its `--prune` would change nothing visible.

## Verification

- `bash .claude/hooks/check-hooks.sh` — ALL CHECKS PASSED (347).
- **Mutation-checked: 24 deliberate breaks, every one turning its own checks
  red.** Each detector, the version sort, the digit filter, the carve-out, the
  worktree keying, the wrapper refusal, the fail-closed path, both `settings.json`
  entries and both arming properties.
- `make test` — 572 passed, 5 xfailed.
- End-to-end against real repository state: the guard refuses a commit in a
  genuinely merged worktree, and is silent in this one and in the main checkout.

## Review found two permitting holes, both fixed in d840e57

Both silent, both in the permitting direction, and the suite was green over
both — the tenth such round on these files.

1. Continuation flags were matched anywhere in a command's arguments, after the
   quotes had been stripped, so `git commit -m "permit rebase --continue"`
   permitted a commit on a merged branch. Matched as the first argument now, and
   never for `commit`. The trade: `git rebase --quiet --continue` is refused.
2. The library probe asked for `cs_split` while the file calls three functions.
   `RAW=$(cs_git_args "$VERB") || continue` cannot tell "not this verb" from "no
   such function", so a half-loaded library permitted every verb through the
   check written to stop exactly that.

## Two things for Bertan

**One acceptance criterion is not met, and cannot be met by an agent.**
`allow_squash_merge` and `allow_rebase_merge` are still enabled on the
repository. The fallback detector rests on *merged implies ancestor*, which is a
property of merge style, so it is unsound until they are disabled: a
squash-merged branch is no ancestor and reads as `ahead > 0`. The gone detector
still catches that branch, which is the point of having two. Changing a
repository setting is not an act an agent takes, so it is left here rather than
done. The header states the dependency and that it is not yet satisfied.

**One interpretation worth a ruling.** The criterion says the guard abstains
when no `refs/remotes/origin/dev-*` exists. That sentence sits in the *active
dev branch* paragraph, so this reads it as governing the ahead/behind fallback
only: the gone detector reads the remote's existence rather than ancestry
against a dev branch, and still refuses in a clone with no dev ref. A branch
whose remote half has been pruned away is merged whether or not this clone has
ever seen a dev branch. Pinned by a check either way, so reversing it is one
edit and one literal.

https://claude.ai/code/session_017GdTHeFNSzxJiZNixD8pFf

### DIFF
--- a/.claude/hooks/check-hooks.sh
+++ b/.claude/hooks/check-hooks.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
-# Regression checks for no-git-push.sh, no-pr-decisions.sh and
-# no-commit-to-main.sh.
+# Regression checks for no-git-push.sh, no-pr-decisions.sh,
+# no-commit-to-main.sh and no-work-on-stale-branch.sh.
 #
 # A hook is a process, so the only way to test one is to run it; what the rule
 # against calling the function under test forbids is deriving the expectation
@@ -158,6 +158,43 @@ says() {  # says <dir> <script> <fragment> <label> <cmd>
          "$label" "$want" "$err"
        FAILED=1 ;;
   esac
+}
+
+# The other half of says: a refusal that must not say something. The fallback
+# detector cannot tell a merged branch from one cut before the dev branch moved,
+# so a message claiming a merge there would be a claim the hook cannot support.
+# Nothing above can catch a message saying too much.
+says_not() {  # says_not <dir> <script> <fragment> <label> <cmd>
+  local dir="$1" script="$2" unwanted="$3" label="$4" cmd="$5" err
+  err=$(printf '%s' "$cmd" | jq -Rs '{tool_name:"Bash",tool_input:{command:.}}' \
+        | ( cd "$dir" && "$HOOKS/$script" ) 2>&1 >/dev/null)
+  case "$err" in
+    *"$unwanted"*) printf '  FAIL %s\n         the refusal must not say |%s|\n         it said |%s|\n' \
+         "$label" "$unwanted" "$err"
+       FAILED=1 ;;
+    *) printf '  ok   says  %s\n' "$label" ;;
+  esac
+}
+
+# A property of a file rather than of a process. See the arming section at the
+# foot of this suite for why one file in .claude/ needs this and the others do
+# not. Fixed strings, not patterns: the expectation is the line as written.
+armed() {  # armed <label> <file> <literal>
+  if grep -qF -- "$3" "$2" 2>/dev/null; then
+    printf '  ok   armed %s\n' "$1"
+  else
+    printf '  FAIL %s\n         expected %s to contain |%s|\n' "$1" "$2" "$3"
+    FAILED=1
+  fi
+}
+
+unarmed() {  # unarmed <label> <file> <literal>
+  if grep -qF -- "$3" "$2" 2>/dev/null; then
+    printf '  FAIL %s\n         %s must not contain |%s|\n' "$1" "$2" "$3"
+    FAILED=1
+  else
+    printf '  ok   armed %s\n' "$1"
+  fi
 }
 
 . ./lib/command-scan.sh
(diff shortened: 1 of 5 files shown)


bgunyel:
## Ruling on the interpretation this PR left open

Bertan's decision: **keep the behaviour as written.** No change to the hook, and
the check that pins it stays as it is.

The criterion reads "the guard abstains when no `refs/remotes/origin/dev-*`
exists". This PR reads that as governing the ahead/behind fallback only, so the
`upstream: gone` detector still refuses in a clone with no dev ref. That reading
stands.

The reason is the narrow one: a branch deleted from the remote was merged, and
that stays true whatever refs this clone happens to hold. The gone detector
never consults a dev branch to reach its conclusion, so nothing about it depends
on one existing.

### A correction to the reasoning, not to the decision

Review argued for this on the grounds that a rotation is exactly when branches
are being merged, so the guard should stay awake through it. **That argument is
wrong and should not be reused.** `branch-hygiene/SKILL.md` pushes `dev-NN+1`
(step 3) before it deletes `dev-NN` (step 4), so a rotation that follows the
procedure never passes through a state with no `origin/dev-*` ref.

The state is reachable, but only narrowly: between the two adjacent lines of
step 3, after `git checkout -b dev-NN+1` and before `git push -u origin
dev-NN+1`; or on a deviation from the procedure that deletes the old branch
first; or in a clone that has never fetched. The hook reads only
`refs/remotes/origin/dev-*` and never a local branch, which is why the local
checkout in step 3 does not close that gap -- and reading a local ref would be
wrong for the reason the push hook already gives, since an agent can create a
branch as readily as it can set configuration.

So the decision holds on the first reason alone. It decides a rarer case than
the PR discussion implied.

### The unmet acceptance criterion is now met

`allow_squash_merge` and `allow_rebase_merge` are both `false` on the
repository, with `allow_merge_commit` and `delete_branch_on_merge` still `true`.
The fallback detector's assumption -- merged implies ancestor -- now holds by
configuration rather than by habit, which is what the ticket asked for.

Measured before the change: a squash-merged worktree branch read as `ahead > 0`
and a commit on it was permitted until the next pruning fetch, at which point
the gone detector caught it. That branch can no longer exist through a pull
request. A squash merge done locally and pushed straight to `dev-NN` would still
break the assumption, but advancing the active dev branch is a reserved act, and
a worktree branch reaches `dev-NN` only by pull request.

PR #64, opened 2026-09-11
Resolve the catch-up ref before permitting it

### LINKED ISSUE #58: Resolve the catch-up ref before permitting it: a diverged local dev-NN retires the fallback detector
## Parent

#36

## Context

Follow-up defect in the guard built by #44, found by a Codex review of `dev-05`
against `main`.

## What is wrong

`no-work-on-stale-branch.sh` establishes staleness against **one ref**:

```sh
DEV=$(git for-each-ref --format='%(refname:short)' 'refs/remotes/origin/dev-*' ... | sort -V | tail -1)
COUNTS=$(git rev-list --left-right --count "$DEV...refs/heads/$CURRENT")
```

`DEV` is `origin/dev-NN`, a remote-tracking ref. `AHEAD == 0` therefore means
this branch is a strict ancestor of **`origin/dev-NN`**, and that — and only
that — is what makes the carve-out's fast-forward claim true.

The carve-out then accepts four spellings by string comparison:

```sh
names_dev() {
  case "$1" in
    "$DEV"|"$DEV_SHORT"|"refs/remotes/$DEV"|"refs/heads/$DEV_SHORT") return 0 ;;
    *) return 1 ;;
  esac
}
```

Two of those four — `dev-NN` and `refs/heads/dev-NN` — name a **different ref**
from the one the ancestry was read against. The header asserts they are the
same:

> DEV is origin/dev-NN; the local branch of the same name, and both branches'
> full ref paths, name the same commit for the purpose of a fast-forward.

Git does not guarantee that. `report-stale-branches.sh` fetches with `--prune`,
which updates `refs/remotes/origin/*` and leaves `refs/heads/*` exactly where
they were. A local `dev-NN` is synchronised only by a checkout-and-pull that
nobody performs in a linked worktree.

## Why it matters

If local `dev-NN` carries commits that `origin/dev-NN` does not, the worktree
branch is an ancestor of `origin/dev-NN` but **not** of local `dev-NN`. Then
`git merge dev-05` is not a fast-forward: it writes a real merge commit onto the
worktree branch. `AHEAD` becomes `> 0`, the branch reads as one carrying work of
its own, and **the fallback detector is retired for that branch permanently** —
the precise outcome the header names as the test for inclusion:

> the test is not "would an agent plausibly write this" but "would this silently
> disable the rule"

Silent, and in the permitting direction. That is the class of defect CLAUDE.md
records nine of in the shared tokeniser alone, every one found by review rather
than by the suite.

This is not an evasion requiring construction. `git merge dev-05` is the
shortest spelling of the command the refusal message itself recommends, and a
local `dev-NN` behind or beside its remote half is the ordinary state of this
repository: worktree pull requests merge into `dev-NN` on GitHub, so
`origin/dev-NN` moves while the local branch sits still.

## The check that is green for the wrong reason

The lifecycle fixture in `check-hooks.sh` creates `refs/remotes/origin/dev-05`,
`origin/dev-foo` and `origin/dev-4` with `update-ref`, and **no local `dev-05`
at all**. So:

```sh
check_in "$WT_STALE" no-work-on-stale-branch.sh ALLOW 'the catch-up merge, short spelling' \
  'git merge dev-05'
```

passes because the hook compares two strings, against a ref that does not exist
in the fixture. Run for real there, that command fails in git — `dev-05`
resolves through `refs/heads/`, `refs/tags/` and `refs/remotes/<name>/`, none of
which matches a remote-tracking `origin/dev-05`. The check is evidence about
string matching and about nothing else.

## What to build

Resolve the supplied ref and require it to name the commit staleness was
measured against. Keep the four-spelling whitelist as the outer gate — it stays
a whitelist, not a rev-parse of arbitrary text — and add the identity check
behind it:

```sh
# Computed once, beside DEV. The ancestry above was read against this commit,
# so this is the commit a catch-up has to be naming for the fast-forward claim
# to hold. A local branch of the same name is a different ref: a pruning fetch
# moves refs/remotes/origin/*, never refs/heads/*.
DEV_OID=$(git rev-parse --verify --quiet "refs/remotes/$DEV^{commit}" 2>/dev/null)

names_dev() {
  local TOK_OID
  case "$1" in
    "$DEV"|"$DEV_SHORT"|"refs/remotes/$DEV"|"refs/heads/$DEV_SHORT") ;;
    *) return 1 ;;
  esac
  [ -n "$DEV_OID" ] || return 1
  TOK_OID=$(git rev-parse --verify --quiet "$1^{commit}" 2>/dev/null) || return 1
  [ "$TOK_OID" = "$DEV_OID" ]
}
```

Costs two `rev-parse` calls, and only on the carve-out path — a merge or rebase
in an already-refused state, which is rare. Everything else still exits before
the library is sourced.

Update the header paragraph quoted above: it currently states the equality as a
fact, and must instead state it as a condition the hook checks.

## The trade this makes, recorded

`git merge dev-05` is **refused** when no local `dev-NN` exists — today it is
permitted. Nothing is lost: run for real, that command fails in git anyway, and
the refusal message already names `origin/dev-05`, the spelling that works. Say
this in the header, in the commit message and in a check, per the repo's rule
that a fix giving up a case says so in all three.

## Not in scope

Widening the whitelist to any revision that happens to resolve to the dev tip —
a raw SHA, `HEAD~3`, another branch parked on the same commit. Such a merge
would be a genuine fast-forward, but the whitelist's job is to be readable by
the next person, and a hook that rev-parses arbitrary command text to decide
whether to permit is a larger claim than this defect needs.

## Fixture work

One fixture cannot hold both states, so build a sibling of the lifecycle
repository with local `dev-05` diverged from `origin/dev-05`. Rejected
alternative: `git branch -f dev-05 <other>` partway down the existing check list
— two lines instead of twenty, but it makes check **order** load-bearing, and an
order-dependent suite is the kind of evidence that quietly stops meaning what it
says.

In the existing lifecycle fixture, create a local `dev-05` at `LIFE_TIP` so the
short-spelling and `refs/heads/` checks stay ALLOW **for the right reason**
rather than by string match.

## Acceptance criteria

- [ ] `DEV_OID` is resolved once from `refs/remotes/$DEV`, beside `DEV`
- [ ] `names_dev` keeps the four-spelling whitelist and additionally requires the
      token to resolve to `DEV_OID`
- [ ] The existing lifecycle fixture gains a local `dev-05` at `LIFE_TIP`; the
      `dev-05` and `refs/heads/dev-05` catch-up checks stay ALLOW
- [ ] A sibling fixture has local `dev-05` diverged from `origin/dev-05`, and in
      it `git merge dev-05` and `git rebase refs/heads/dev-05` are BLOCK while
      `git merge origin/dev-05` stays ALLOW
- [ ] A check covers no local `dev-NN` at all: the short spelling is BLOCK, and
      the header, the commit message and that check all record the trade
- [ ] A check covers `DEV_OID` unresolvable: the carve-out is not taken
- [ ] Mutation-checked — reverting `names_dev` to the plain string comparison
      turns the new checks red, and no fix lands without a check that fails
      without the fix
- [ ] The header paragraph asserting the four spellings name the same commit is
      rewritten to state the condition the hook now checks

bgunyel:
Closes #58.

## What was wrong

The carve-out in `no-work-on-stale-branch.sh` accepted four spellings of the
active dev branch by string comparison. Two of them — `dev-NN` and
`refs/heads/dev-NN` — name a different ref from the one staleness was measured
against. `AHEAD == 0` says this branch is a strict ancestor of
`refs/remotes/origin/dev-NN`, and that alone is what makes the fast-forward
claim true. The header asserted the four name the same commit; git does not
guarantee it, and here they routinely disagree — worktree pull requests merge
into `dev-NN` on GitHub, so `origin/dev-NN` moves while the local branch sits
still, and `report-stale-branches.sh` fetches with `--prune`, which updates
`refs/remotes/origin/*` and leaves `refs/heads/*` exactly where they were.

Against a local `dev-NN` that has forked, `git merge dev-NN` — the shortest
spelling of the command the refusal message itself recommends — writes a real
merge commit. `AHEAD` becomes `> 0`, the branch reads as one carrying work of
its own, and the fallback detector is retired for that branch permanently.
Silent, and in the permitting direction.

## What changed

`names_dev` keeps the four-spelling whitelist as the outer gate and now resolves
the token behind it, requiring it to name `DEV_OID`, the commit the ancestry was
read against. It still rev-parses four fixed strings and never arbitrary command
text — widening the whitelist to any revision resolving to the dev tip was out
of scope and stayed out.

The header paragraph asserting the equality is rewritten to state it as a
condition the hook checks.

## The trade, recorded

Where no local `dev-NN` exists at all, `git merge dev-NN` is now **refused**; it
used to be permitted. Nothing is lost: run for real in that state the command
fails in git anyway, since `dev-NN` resolves through `refs/heads/`, `refs/tags/`
and `refs/remotes/<name>/`, and a remote-tracking `origin/dev-NN` is none of
those. The refusal already names `origin/dev-NN`, which works.

Said in the file header, in the commit message, and in a check of its own.

## Fixtures

The lifecycle fixture had no local `dev-05` at all, so the short-spelling checks
were green for the wrong reason — two strings compared against a ref that did
not exist there, for a command that fails in git if run. It gains a local
`dev-05` at `LIFE_TIP`, and three siblings hold the other states: forked local
`dev-05`, no local `dev-05`, and a dev tip that is not a commit. Separate
repositories rather than `git branch -f` partway down the list, which would make
check **order** load-bearing.

Diverged there means **forked**, not merely ahead: a local branch one commit
ahead of `origin/dev-NN` still has the worktree branch as an ancestor, so the
merge would fast-forward and the fixture would prove only that two strings
differ. The first draft of this branch made exactly that mistake and the review
caught it; the fixture now asserts both halves of the shape it needs.

## Mutation checks

- reverting `names_dev` to the plain string comparison → the four new BLOCK
  checks go red
- dropping the `DEV_OID` emptiness guard → its `armed` line goes red
- removing the local `dev-05` from the lifecycle fixture → the two
  short-spelling ALLOWs go red, which is what shows they now pass for the right
  reason
- rebuilding the forked fixture linearly → the suite stops at its own guard

## Two deviations from the issue, for review

**`DEV_OID` placement.** The issue says "resolved once from `refs/remotes/$DEV`,
beside `DEV`". It is beside `DEV_SHORT` instead, 87 lines later. The issue's own
cost claim — "only on the carve-out path … everything else still exits before
the library is sourced" — and its snippet comment "the ancestry **above**" both
point there, and the common path exits before it. Argued in the file.

**The unresolvable-`DEV_OID` check is reframed, not skipped.** That line cannot
be reached by running the hook: `DEV` is the short name of the very ref
`DEV_OID` is read from, so a dev tip that will not resolve is one `git rev-list`
cannot read either, and the file abstains at `STATE=CLEAR` before the carve-out
is considered. It is checked two ways instead, and the suite says which is
which — the abstention driven as a process, the guard behind it pinned as a
property of the file. The guard is kept because a comparison against an empty
string is not a comparison: drop the line and an empty `TOK_OID` would equal an
empty `DEV_OID`, permitting the merge on exactly the state that could not be
read.

## Verification

`bash .claude/hooks/check-hooks.sh` passes — 562 checks. Both files pass
`bash -n`. No dependency changed, so `make upgrade-safe` has not been run; it is
the floor before this closes.

`make test` is 571 passed / 1 failed on this branch, and that failure is
pre-existing environment drift — `alembic` and `mako` from the `migrations`
group are not installed in `.venv`, repaired with `uv sync --all-groups`. This
diff touches no Python.

https://claude.ai/code/session_019HkmeYcZLfPsF22F4L4Gp8

### DIFF
--- a/.claude/hooks/check-hooks.sh
+++ b/.claude/hooks/check-hooks.sh
@@ -197,8 +197,16 @@ says_not() {  # says_not <dir> <script> <fragment> <label> <cmd>
 # A property of a file rather than of a process. See the arming section at the
 # foot of this suite for why one file in .claude/ needs this and the others do
 # not. Fixed strings, not patterns: the expectation is the line as written.
+# Comments are stripped before the search, and that is the whole point rather
+# than a detail: `grep -qF` over raw file text matches a literal that has been
+# commented OUT, so every one of these stayed green while the line it names did
+# nothing. Found by review of PR #64, not by this suite. Comment-out-and-leave
+# is how a shell script ordinarily gets edited, not a constructed evasion, so
+# this is the permitting-direction defect these checks exist to catch, in the
+# checks themselves. None of the literals below contains a `#`, so stripping
+# from the first one is safe for them.
 armed() {  # armed <label> <file> <literal>
-  if grep -qF -- "$3" "$2" 2>/dev/null; then
+  if sed 's/[[:space:]]*#.*$//' "$2" 2>/dev/null | grep -qF -- "$3"; then
     printf '  ok   armed %s\n' "$1"
   else
     printf '  FAIL %s\n         expected %s to contain |%s|\n' "$1" "$2" "$3"
@@ -1732,7 +1740,53 @@ check_in "$WT_NOLOC" no-work-on-stale-branch.sh BLOCK 'no local dev-05: the shor
 check_in "$WT_NOLOC" no-work-on-stale-branch.sh BLOCK 'no local dev-05: refs/heads/dev-05 names nothing either' \
   'git merge refs/heads/dev-05'
 
-# STATE THREE: the dev tip does not resolve to a commit. The hook holds
+# STATE THREE: a local dev-05 merely BEHIND origin/dev-05 -- not forked, just
+# not pulled. THE SECOND HARMLESS CASE THIS FIX GIVES UP, and the one that
+# starts firing as soon as it lands: worktree pull requests merge into dev-05 on
+# GitHub, so origin/dev-05 moves while the local branch sits still, and that is
+# the ordinary state of this repository rather than an edge of it.
+#
+#   root ---- beh-stale, dev-05 ---- origin/dev-05
+#
+# From beh-stale, `git merge dev-05` is `Already up to date.` -- it writes
+# nothing and masks nothing. It is refused anyway, because the test is an
+# identity and the local branch is not origin/dev-05. Separating this case from
+# the forked one means asking about ancestry, and a hook that rev-parses its way
+# to a merge-base decision is a larger claim than this defect needs. Recorded
+# here, in the file header and in the commit message, per the repository's rule
+# that a fix giving up a case says so in all three.
+BEH="$FIXTURES/behind-dev"
+git init -q -b main "$BEH"
+GH_="git -C $BEH -c user.email=checks@example.invalid -c user.name=checks"
+$GH_ remote add origin "$FIXTURES/unreachable-remote.git"
+$GH_ commit -q --allow-empty -m root
+$GH_ commit -q --allow-empty -m "where local dev-05 stopped"
+BEH_LOCAL=$($GH_ rev-parse HEAD)
+$GH_ commit -q --allow-empty -m "where origin/dev-05 went without it"
+$GH_ update-ref refs/remotes/origin/dev-05 "$($GH_ rev-parse HEAD)"
+$GH_ branch dev-05 "$BEH_LOCAL"
+$GH_ branch beh-stale "$BEH_LOCAL"
+$GH_ worktree add -q "$BEH/wt-beh" beh-stale
+WT_BEH="$BEH/wt-beh"
+need_worktree "$WT_BEH" behind-dev
+# Behind, not forked: the worktree branch IS an ancestor of the local branch, so
+# the merge given up here is a no-op rather than a merge commit. That is the
+# difference from STATE ONE, and asserting it is what stops this fixture
+# quietly turning into a copy of that one.
+$GH_ merge-base --is-ancestor refs/heads/beh-stale refs/heads/dev-05 || {
+  echo "beh-stale is not an ancestor of local dev-05, so this is the forked case again; the checks below prove nothing" >&2
+  exit 1
+}
+$GH_ merge-base --is-ancestor refs/heads/dev-05 refs/remotes/origin/dev-05 || {
+  echo "local dev-05 is not behind origin/dev-05; the checks below prove nothing" >&2
+  exit 1
+}
+check_in "$WT_BEH" no-work-on-stale-branch.sh ALLOW 'local dev-05 behind: the remote spelling is the catch-up' \
+  'git merge origin/dev-05'
+check_in "$WT_BEH" no-work-on-stale-branch.sh BLOCK 'local dev-05 behind: a harmless no-op merge, refused, and recorded as given up' \
+  'git merge dev-05'
+
+# STATE FOUR: the dev tip does not resolve to a commit. The hook holds
 # `[ -n "$DEV_OID" ] || return 1` for it, and that line cannot be reached by
 # running the hook: DEV is the short name of the very ref DEV_OID is read from,
 # so a dev tip that will not resolve is one `git rev-list` cannot read either,
@@ -1896,11 +1950,15 @@ armed 'and the report derives it identically' \
 # The carve-out's identity test, pinned as three lines rather than driven as a
 # process. Two of them are driven, by the diverged and no-local fixtures above;
 # the third -- an unresolvable dev tip -- is not reachable by running this hook,
-# because the ancestry read fails first and the file abstains. It is kept
-# because the carve-out is the only permitting path out of a refused state, and
-# a comparison against an empty string is not a comparison at all: drop the
-# line and `TOK_OID` empty would equal `DEV_OID` empty, permitting the merge on
-# exactly the state that could not be read. This says so; nothing above can.
+# because the ancestry read fails first and the file abstains.
+#
+# The guard line is also redundant with the comparison that follows it, and the
+# first version of this suite claimed otherwise -- that dropping it would let an
+# empty TOK_OID equal an empty DEV_OID. That was false: the rev-parse above the
+# comparison returns on failure and prints an OID on success, so TOK_OID is
+# never empty there. What these three lines pin is that the identity test is
+# spelled the way the file says it is; they are not evidence that any one of
+# them decides an outcome, and the middle one does not.
 armed 'the dev tip is resolved to a commit, from the ref the ancestry was read against' \
   "$HOOKS/no-work-on-stale-branch.sh" \
   'DEV_OID=$(git rev-parse --verify --quiet "refs/remotes/$DEV^{commit}" 2>/dev/null)'
--- a/.claude/hooks/check-hooks.sh
+++ b/.claude/hooks/check-hooks.sh
@@ -206,6 +206,17 @@ armed() {  # armed <label> <file> <literal>
   fi
 }
 
+# A fixture guard rather than a check, and it stops the suite rather than
+# failing one line. An unmade worktree makes ( cd "$dir" && hook ) return 1,
+# which reads as ALLOW -- so every ALLOW-expecting check against it would pass
+# without the hook ever running. That is the shape this suite exists to not
+# have, so it is said once here rather than three times below.
+need_worktree() {  # need_worktree <dir> <fixture name>
+  [ -d "$1" ] && return 0
+  echo "the $2 worktree was not created; the checks against it prove nothing" >&2
+  exit 1
+}
+
 unarmed() {  # unarmed <label> <file> <literal>
   if grep -qF -- "$3" "$2" 2>/dev/null; then
     printf '  FAIL %s\n         %s must not contain |%s|\n' "$1" "$2" "$3"
@@ -1452,6 +1463,12 @@ LIFE_TIP=$($GL rev-parse HEAD)
 $GL update-ref refs/remotes/origin/dev-05 "$LIFE_TIP"
 $GL update-ref refs/remotes/origin/dev-foo "$LIFE_BASE"
 $GL update-ref refs/remotes/origin/dev-4 "$LIFE_BASE"
+# A local dev-05 at the same commit, which is the state a checkout-and-pull
+# leaves. Without it the short-spelling and refs/heads/ catch-up checks below
+# were green for the wrong reason: the hook compared two strings against a ref
+# that did not exist here at all, and `git merge dev-05` run for real in this
+# fixture fails in git. The sibling fixtures hold the other two states.
+$GL branch dev-05 "$LIFE_TIP"
 
 # ahead == 0, behind == 1. The fallback detector's case, and nothing else: this
 # branch has no upstream configured, so `[gone]` cannot be what refuses it.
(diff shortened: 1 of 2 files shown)


bgunyel:
## Review of `c7accb8`

Ran the suite and reproduced every mutation claim independently, on throwaway copies of `.claude/` — nothing in the branch was touched.

### Acceptance criteria (#58)

| # | Criterion | Status |
|---|---|---|
| 1 | `DEV_OID` resolved once from `refs/remotes/$DEV`, **beside `DEV`** | Deviated, disclosed — see F3 |
| 2 | `names_dev` keeps the whitelist, adds the identity test | ✅ |
| 3 | Lifecycle fixture gains local `dev-05` at `LIFE_TIP`; both spellings stay ALLOW | ✅ (and a `refs/heads/dev-05` ALLOW that did not exist before) |
| 4 | Sibling fixture diverged: two BLOCK, `origin/` spelling ALLOW | ✅ |
| 5 | No-local-`dev-NN` check; trade in header + commit message + check | ✅ but under-stated — see F4 |
| 6 | A check covers `DEV_OID` unresolvable: the carve-out is not taken | ❌ **not met** — see F1/F2 |
| 7 | Mutation-checked | ✅ all four claims reproduce exactly |
| 8 | Header paragraph rewritten as a condition, not a fact | ✅ |
| 9 | `check-hooks.sh` passes | ✅ 562 checks, 0 fails |

Mutations rerun: reverting `names_dev` to the plain string comparison reddens exactly the four new BLOCKs; removing `$GL branch dev-05` from the lifecycle fixture reddens exactly the two short-spelling ALLOWs; deleting the `DEV_OID` guard reddens its `armed` line; rebuilding the fork linearly stops the suite at its own `div-stale is an ancestor of local dev-05` guard, exit 1.

The core fix is correct and the evidence behind it is real. The findings below are all in the periphery.

### F1 — the stated reason for the `DEV_OID` emptiness guard is false (medium)

The file header, the suite comment, the commit message and the PR body all carry the same claim:

> drop the line and `TOK_OID` empty would equal `DEV_OID` empty, permitting the merge on exactly the state that could not be read

`TOK_OID` can never be empty at that comparison. The preceding line is `TOK_OID=$(git rev-parse …) || return 1` — if the rev-parse fails the function has already returned, and if it succeeds it printed an OID. Running the function body with the guard deleted and `DEV_OID=""`: all four whitelisted spellings are still refused, plus a non-whitelisted control. The guard cannot change an outcome.

It is harmless belt-and-braces, but the justification is asserted rather than true — in a file whose whole argument is that an asserted equality is what caused this defect to begin with.

### F2 — AC 6 rests on the seam that is mutation-blind (medium)

The three new `armed` lines are `grep -qF` over file text. Commenting the guard out rather than deleting it:

```sh
  # disabled: [ -n "$DEV_OID" ] || return 1
```

gives **562 checks, ALL CHECKS PASSED**. The literal survives inside the comment. This is the same shape as the `--prune` literal in the arming section, reproduced in a new place — and comment-out-and-replace is the ordinary way a shell script gets edited, not a constructed evasion.

Taken with F1, AC 6 is unmet three ways over: the named state is unreachable, the guard behind it is inert, and the check pinning it cannot detect its removal.

On unreachability — the PR's claim survives an attempt to break it. A local branch literally named `origin/dev-05` was the promising route, since `origin/dev-05` resolves through `refs/heads/` before `refs/remotes/`, which would let `git rev-list` succeed while `refs/remotes/origin/dev-05` pointed at a blob. It does not work: `for-each-ref --format='%(refname:short)'` emits `remotes/origin/dev-05` under that ambiguity, the `^origin/dev-[0-9]+$` filter rejects it, `DEV` is empty and the file abstains. The claim holds; it is the guard's stated purpose that does not.

### F3 — the cost claim behind the AC 1 deviation is inexact (low)

> one rev-parse here … paid only on the carve-out path -- a merge or rebase in an already-refused state, which is rare

`DEV_OID=` sits at line 276 and is unconditional once `STATE != CLEAR`, so `git status` in a stale worktree pays it too. Irrelevant against a five-second budget, but this is the argument used to justify moving `DEV_OID` away from where the issue placed it, so it should be exact: paid once per command in a refused state, and one more per whitelisted token.

### F4 — the recorded trade is narrower than the trade made (low)

Recorded: no local `dev-NN` ⇒ refused. Actual: **any** local `dev-NN` whose OID differs from `origin/dev-NN`'s.

The commonest such state is the one this PR's own reasoning describes — local `dev-05` merely *behind* `origin/dev-05` after worktree pull requests land on GitHub. Built as a fixture: `behind/ahead = 1 0`, and `git merge dev-05` there reports `Already up to date.` and writes nothing. The hook now returns 2 on it.

Safe direction, and the refusal names the spelling that works. But it is a second harmless case given up, and by this repository's own rule — a fix giving up a case says so in the header, the commit message and a check — it is recorded in none of the three and has no check. It starts firing as soon as this lands and the next pull request moves `origin/dev-05`.

bgunyel:
All four addressed in `4eebeab`. Two of them were claims this branch made rather than things it did, which is the part worth saying plainly.

### F1 — you're right, and the justification was the same sin the file argues against

Confirmed by reading: `TOK_OID=$(git rev-parse …) || return 1` means the function has already returned if the rev-parse failed, and printed an OID if it succeeded. `TOK_OID` is never empty where the comparison runs, so an unreadable `DEV_OID` already fails that comparison. The guard cannot change an outcome.

The line is **kept** — #58 specifies it, and deleting it would be a third deviation from the issue on top of the two already open — but it is now labelled as redundant at the point of the read, and the comparison below it is marked as the thing that actually refuses. Header, suite comment and commit message all corrected; the PR body above still carries the old claim and is wrong where it says so.

If you'd rather have the line gone than labelled, say so and it goes — it is two lines of diff either way.

### F2 — fixed, and it reached further than the new checks

`armed()` now strips comments before the search. Your repro reproduces: commenting the guard out rather than deleting it left the suite green at 562. With the fix, that mutation reddens the line it names.

I ran it on a throwaway copy of `.claude/` rather than in place — the harness refused to let me comment out a guard line in the live hook, which is the correct instinct even though it was a mutation test.

This also covers the pre-existing `--prune` and dev-derivation literals, which had the same shape. All of them still pass, and none of the pinned literals contains a `#`, so stripping from the first one is safe for them. Noted in the helper.

### F3 — corrected to be exact

The header now says: one rev-parse once per command in a refused state — `git status` in a stale worktree pays it too, not only a merge — and one more per whitelisted token inside `names_dev`. Since that cost is the whole argument for moving `DEV_OID` away from where #58 put it, it should be exact.

### F4 — this was the one I'd got substantively wrong

Recorded trade was "no local `dev-NN`". Actual trade is "any local `dev-NN` whose OID differs", and the commonest member of that set is the merely-*behind* branch — the ordinary state here, not an edge, and it starts firing the moment `origin/dev-05` next moves.

Now a fixture of its own (`behind-dev`), asserting both halves of the shape so it cannot quietly become a copy of the forked one:

```
root ---- beh-stale, dev-05 ---- origin/dev-05
```

`git merge dev-05` there is `Already up to date.` and is refused; `git merge origin/dev-05` stays ALLOW. Recorded in the header, in the commit message and in that check, per the rule. The reason it is given up rather than fixed is in the header too: separating it from the forked case means asking about ancestry, and a hook that rev-parses its way to a merge-base decision is a larger claim than this defect needs.

### AC 6

I'm not claiming it met. The state is unreachable, the guard behind it is inert, and — until F2 — the check pinning it was blind to its own removal. Two of those three are now true statements in the file rather than false ones, and the third is fixed. What the three `armed` lines pin is that the identity test is spelled the way the file says; they are not evidence that any one line decides an outcome, and the suite now says that instead of the opposite.

If AC 6 should be met rather than re-described, the honest route is to drop the criterion in #58 as written — the state it names cannot be constructed against this hook — and I'd rather you make that call than quietly close it as done.

### Mutations, rerun

- `names_dev` → plain string comparison: **five** BLOCKs red now, the new behind case among them
- emptiness guard commented out: its `armed` line red (was invisible)
- local `dev-05` removed from the lifecycle fixture: the two short-spelling ALLOWs red
- fork rebuilt linearly: suite stops at its own guard, exit 1

`bash .claude/hooks/check-hooks.sh` — 564 checks, all pass.

`make upgrade-safe` still outstanding before this closes; no dependency changed.

https://claude.ai/code/session_019HkmeYcZLfPsF22F4L4Gp8

PR #67, opened 2026-09-11
Pin the whole active-dev derivation, not its tail

### LINKED ISSUE #62: Pin the whole active-dev derivation, not its tail: the unpinned half is the half that moves
## Parent

#36

## Context

Found by a two-axis review of `dev-05` against `main` (standards axis). The
review first read this as plain duplicated code; it is not, and the real defect
is narrower and worse.

## What is not wrong

The duplication is deliberate and its reason is recorded at
`check-hooks.sh:1720-1728`:

> the guard must read its state before it may depend on `lib/` at all. That is
> what lets it fail closed only on a branch it has an opinion about, rather
> than refusing every command in every worktree whenever a library is missing.
> The cost of that ordering is two copies, so the copies are pinned instead of
> shared.

That argument holds. **This ticket does not propose extracting the derivation
into `lib/`.** It proposes fixing the pin that was accepted in its place.

## What is wrong

The mitigation for the two copies is two `armed` checks against one literal:

```sh
DEV_DERIVATION="| grep -E '^origin/dev-[0-9]+\$' | sort -V | tail -1)"
armed 'the guard derives the active dev branch this way' \
  "$HOOKS/no-work-on-stale-branch.sh" "$DEV_DERIVATION"
armed 'and the report derives it identically' \
  "$HOOKS/report-stale-branches.sh" "$DEV_DERIVATION"
```

The derivation is two lines. The pin covers the second:

```sh
DEV=$(git for-each-ref --format='%(refname:short)' 'refs/remotes/origin/dev-*' 2>/dev/null \
      | grep -E '^origin/dev-[0-9]+$' | sort -V | tail -1)
```

Everything before the pipe — the command, the `--format`, and the ref glob
`'refs/remotes/origin/dev-*'` — is pinned in neither file. Change the glob in
one file and not the other, or change `--format` to `%(refname)` in one, and
both checks still pass while the two files derive different things.

Two smaller holes in the same check:

- **`armed` is `grep -qF` for a literal anywhere in a file.** It asserts each
  file contains the constant; it never compares the two files to each other. If
  a third derivation were added to either file, the check would be satisfied by
  the old one.
- **The rationale is written twice, in different words.** The version sort and
  the digit filter are argued at `no-work-on-stale-branch.sh:45-49` and again at
  `report-stale-branches.sh:71-75`. Nothing pins those to each other either, so
  a future correction to one leaves the other asserting the superseded reason —
  and CLAUDE.md treats that as a defect in its own right:

  > a stale docstring here is a defect

## Why it matters

The comment that accepts the duplication states the consequence of a
divergence, and it is the permitting direction:

> A divergence here silently unarms the guard or misreports the branch.

The glob is the half most likely to move. `dev-*` is the naming convention this
repository is currently mid-argument with — `origin/dev-05-backup` and
`origin/dev-foo` are both named in the existing rationale as shapes the filter
exists to exclude — so a future change to which refs count is a change to the
glob, which is exactly the unpinned half. A guard that stops finding the active
dev branch abstains rather than refuses, so the failure is silent.

This is the ninth-defect shape CLAUDE.md records, applied to the check suite
itself rather than to a hook:

> A check suite is evidence about the cases it names and about nothing else.

Here it names the tail of a pipeline and is read as evidence about the pipeline.

## Suggested fix

Pin the whole derivation, not its tail — both lines as one literal, including
the glob and the `--format` — and add a check that the two files' derivations
are equal to each other rather than each equal to a constant. Either extend
`DEV_DERIVATION` to the full two-line string, or extract both derivations by
line range and compare them.

Per the standard this repository holds guard code to, the fix lands with a check
that fails without it: change the glob in one file only and confirm the new
check goes red while the two existing `armed` checks stay green.

## Evidence

- `.claude/hooks/check-hooks.sh:1720-1733` — the comment, the literal, the two checks
- `.claude/hooks/check-hooks.sh:200-207` — `armed` is `grep -qF`
- `.claude/hooks/no-work-on-stale-branch.sh:150-151` and `report-stale-branches.sh:77-78` — the two copies
- `.claude/hooks/no-work-on-stale-branch.sh:45-49` and `report-stale-branches.sh:71-75` — the rationale, twice

bgunyel:
Closes #62.

## The defect, measured

The active dev branch is derived by hand in two files, and the duplication is
deliberate — the guard must read its state before it may depend on `lib/` at
all. What paid for it was a pin, and the pin covered the second of the two
lines:

    DEV_DERIVATION="| grep -E '^origin/dev-[0-9]+\$' | sort -V | tail -1)"

Everything left of the first pipe — the command, the `--format`, and the ref
glob `'refs/remotes/origin/dev-*'` — was pinned in neither file. Changing the
glob in `report-stale-branches.sh` alone left the **whole suite green**,
including the check labelled *"and the report derives it identically"*. A guard
that stops finding the active dev branch abstains rather than refuses, so that
divergence would have been silent.

## Why not just extend the literal

`armed` is `grep -qF`, and grep reads a pattern containing a newline as two
patterns: a two-line fixed string is satisfied by a file holding **either line
alone**. Measured on a two-line fixture. That is why the fix extracts and
compares rather than greps.

## What lands

Both derivations are extracted by content — not by line range, which goes stale
silently the moment either file gains a line above it — and checked three ways:
each file holds exactly one, the two are equal to each other, and each is the
derivation as pinned, glob and `--format` included. The counts are what close
the case: two files that both extract to nothing are equal to each other.

The rationale for the filter and the version sort was also written twice, in
different words, with nothing holding those to each other. It is argued once
now, in `no-work-on-stale-branch.sh`'s header — which takes over the two details
only the report had — and each file carries the same one-line pointer, pinned in
both, in place of its own copy.

## Evidence

Eight mutations, each turning red for its own reason:

| mutation | what goes red |
|---|---|
| glob changed in the report only | equality, the report's pin |
| `--format` changed in the guard only | equality, the guard's pin |
| glob changed identically in both | both pins; equality stays green |
| a second derivation in the report | that file's count, equality |
| neither file matching the anchor | both counts — equality alone passes |
| the pairing pointer dropped | that file's pointer check |
| the pointer moved to the file head | `beside` — `armed` passed this |
| the guard's argument dropped | the argument pin |

The second commit fixes four findings from the two-axis review of the first,
one of them the ticket's own defect shape appearing in the checks that fix it:
the pointer checks were labelled *"beside its derivation"* while `armed` looks
anywhere in the file.

Suite: 556 checks, all pass. `make test` is unchanged by this branch (571
passed, 5 xfailed; the one failure is pre-existing venv drift —
`alembic`/`mako` from the `migrations` group are not installed — and this branch
touches no Python).

## What these checks do not see

Named where they are defined. The count finds readers spelled with
`for-each-ref`, so one written as `git branch -r` would read the same refs
uncounted; and it counts matching lines anywhere in a file, comments included,
so quoting the pipeline in either hook's header turns the suite red although
nothing changed.

https://claude.ai/code/session_01NNtmv9SCTyzRZT3E8J13Km

### DIFF
--- a/.claude/hooks/check-hooks.sh
+++ b/.claude/hooks/check-hooks.sh
@@ -234,6 +234,10 @@ need_worktree() {  # need_worktree <dir> <fixture name>
 # line rather than a trailing remark. Routed through `armed`, such a pin cannot
 # pass: measured, not reasoned, on the boundary-section check that dev-05 is
 # red on today. This reads the file as written, and the name says which.
+prose_count() {  # prose_count <file> <literal> -- how many lines say it
+  grep -cF -- "$2" "$1" 2>/dev/null
+}
+
 written() {  # written <label> <file> <literal> -- the file as written, # and all
   if grep -qF -- "$3" "$2" 2>/dev/null; then
     printf '  ok   written %s\n' "$1"
@@ -2125,13 +2129,34 @@ beside 'the guard names the pairing beside its derivation' \
 beside 'and the report names it identically' \
   "$HOOKS/report-stale-branches.sh" "$PAIRING"
 # A pointer to an argument is worth what the argument is worth, and the report's
-# now points at prose in another file. Nothing kept that prose alive, so delete
-# the guard's header and the report goes on citing a reason that is no longer
-# written anywhere -- the stale-docstring defect moved rather than fixed. This
-# pins the claim itself, both halves of it, in the one place it is now made.
+# now points at prose in another file. Two ways that goes wrong, and the second
+# is the one this branch would otherwise have left open.
+#
+# Delete the guard's header and the report cites a reason no longer written
+# anywhere: the stale-docstring defect moved rather than fixed. Re-add a second
+# copy to the report and the defect is back exactly as it was -- two arguments,
+# nothing holding them to each other -- which is what the report's own new
+# comment says is wrong: "a second copy of an argument goes stale in silence
+# when the first one is corrected." The code duplication got a count for that
+# reason; the prose duplication fixed in the same breath did not, and the
+# asymmetry was the gap. Both directions are counted now, the same shape as the
+# two counts above, and both halves of the claim are in the literal.
+#
+# The limit is the one the counts above have: this finds the argument as
+# written, so a reworded second copy is a second copy uncounted.
 ARGUMENT='sort makes dev-09 beat dev-10, and an unfiltered glob lets origin/dev-foo'
-written 'the argument the report points at is still made in the guard' \
-  "$HOOKS/no-work-on-stale-branch.sh" "$ARGUMENT"
+tok 'the argument the report points at is made in the guard, once' \
+    '1' "$(prose_count "$HOOKS/no-work-on-stale-branch.sh" "$ARGUMENT")"
+tok 'and the report does not argue it a second time' \
+    '0' "$(prose_count "$HOOKS/report-stale-branches.sh" "$ARGUMENT")"
+# And the sentence that does the pointing: without it the report holds a bare
+# pairing pointer and no trace of where its reasoning went. `beside` rather than
+# `written`, because a pointer that is not beside the derivation is not doing
+# the job the pointer exists for -- the same standard the pairing line is held
+# to four lines above.
+POINTER="no-work-on-stale-branch.sh's header, rather than twice here in different words"
+beside 'the report says where the argument was moved to' \
+  "$HOOKS/report-stale-branches.sh" "$POINTER"
 
 # The carve-out's identity test, pinned as three lines rather than driven as a
 # process. Two of them are driven, by the diverged and no-local fixtures above;
--- a/.claude/hooks/check-hooks.sh
+++ b/.claude/hooks/check-hooks.sh
@@ -224,12 +224,15 @@ unarmed() {  # unarmed <label> <file> <literal>
 # stale the moment either file gains a line above it, and goes stale silently --
 # and on what the derivation reads rather than on the name it assigns, so a
 # second reader of those refs is counted rather than hidden behind the first.
-# What that counts is readers spelled with for-each-ref: one written as
-# `git branch -r` would read the same refs and not be counted, and the suite
-# says nothing about it. The count is here because equality alone does not
-# close the case -- two files that both extract to nothing are equal, so the
-# three checks are needed together and each was mutated to confirm it.
-dev_reads() {  # dev_reads <file> -- how many lines read the dev refs
+#
+# Two limits, both named because a check is evidence about what it names. The
+# count finds readers spelled with for-each-ref, so one written as
+# `git branch -r` would read the same refs uncounted -- the permitting
+# direction, and the reason the pin below is there too. And it counts matching
+# lines anywhere in the file, comments included, so quoting the pipeline in
+# either hook's header turns the suite red although nothing has changed: the
+# refusing direction, visible and one edit away.
+dev_read_count() {  # dev_read_count <file> -- how many lines read the dev refs
   grep -cE 'for-each-ref.*refs/remotes/origin/dev-' "$1" 2>/dev/null
 }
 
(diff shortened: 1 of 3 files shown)


bgunyel:
## Review: acceptance criteria met; one blocker before this can land

Verified by running the suite and reproducing every mutation, not by reading the diff.

### The ticket's criteria are satisfied

#62 asked for three things and each is here:

1. The whole derivation is pinned, glob and `--format` included, as one heredoc literal.
2. The two files are compared to each other, not each to a constant — plus two "exactly one reader" counts.
3. It lands with a check that fails without it.

I confirmed the premise as well as the fix. On `dev-05`, changing the glob in `report-stale-branches.sh` alone leaves the suite **entirely green — 564 checks, 0 failures — including the one labelled "and the report derives it identically."** The same mutation on this branch turns two checks red. The defect was real and this closes it.

All eight rows of the evidence table reproduce exactly as tabulated. Two claims I measured rather than accepted:

- **The `grep -qF` argument is correct.** A two-line fixed-string pattern matches a file holding *either line alone* — confirmed on three fixtures. Extending the old literal would have been a non-fix, so extract-and-compare was the right call rather than a stylistic one.
- **The counts are load-bearing, exactly as claimed.** With both anchors renamed away, both counts go red while equality stays *green*. Two files that extract to nothing are equal to each other; equality alone would not have closed it.

Check count (556) is accurate. `make test` is untouched by scope — three `.sh` files, no Python.

### Blocker: merging this turns `dev-05` red

This branch was cut from `87b4e6b`. `dev-05` has since advanced by #64, which touches **the same two files**. The merge is textually clean and semantically broken.

`4eebeab` redefined `armed` to strip comments before grepping — deliberately, so that commenting a line out can no longer satisfy a pin. Its own header records the invariant: *"None of the literals below contains a `#`."* `armed` is now a code-only pin by design.

The new `ARGUMENT` check uses `armed` to pin prose that exists **only inside a comment** — the guard's header paragraph. Under the new `armed` it can never pass:

```
FAIL the argument the report points at is still made in the guard
     expected .../no-work-on-stale-branch.sh to contain |sort makes dev-09 beat dev-10, and an unfiltered glob lets origin/dev-foo|
```

Merged result: 570 checks, 1 FAIL, exit 1. The `PAIRING` checks survive only because they route through this branch's own `dev_pointer`/`beside` rather than through `armed`.

This is the ticket's own defect shape, one level up: a helper changed meaning underneath a check, and the check's label goes on claiming what it no longer tests.

**Fix, validated:** give the prose pin a comment-reading helper of its own instead of `armed` — the raw file is the point, so say so where it is defined.

```sh
argued() {  # argued <label> <file> <literal> -- prose, so the raw file
  if grep -qF -- "$3" "$2" 2>/dev/null; then
    printf '  ok   argued %s\n' "$1"
  else
    printf '  FAIL %s\n         expected %s to still argue |%s|\n' "$1" "$2" "$3"
    FAILED=1
  fi
}
```

Merged plus that change: 570 checks, all pass, and the check still goes red when the argument prose is dropped, so it keeps its mutation coverage.

### Two residual gaps, neither in the stated criteria

- The report's pointer sentence into the guard's header can be deleted with nothing going red. That leaves an *absent* rationale rather than a stale one, so it is milder than the defect this fixes.
- **The third defect can silently return.** Re-adding a second copy of the rationale to the report is uncaught. The "count the readers" discipline was applied to the code duplication via `dev_read_count` but not to the prose duplication fixed in the same commit — although the report's new comment argues precisely why it matters: *"a second copy of an argument goes stale in silence when the first one is corrected."* The symmetric check would be a count, the same shape as the two already here.

https://claude.ai/code/session_01QJy3KQup8qNPnB9Ew5LSP2

bgunyel:
Blocker fixed, both residual gaps closed — and the same defect was already merged on `dev-05`.

## The blocker, and one you did not have yet

You were right, and the fix is in `39fca21`. But `dev-05` had advanced past the point you reviewed: it is now at `7cb4891` (#64, #65, #66), and **it is red on its own**, before this branch is involved:

```
$ git archive origin/dev-05 | tar -x -C /tmp/dev05 && cd /tmp/dev05
$ bash .claude/hooks/check-hooks.sh
  FAIL the extracted section is the boundary section
SOME CHECKS FAILED
```

#66's boundary-section check pins a markdown heading — `## What an unattended agent may do to this repository` — through `armed`. Markdown opens a heading with the character `armed` now cuts on, and the pattern is `[[:space:]]*#.*$`, so the *whole line* is erased, not a trailing remark. That check has been red since #66 merged, for the same reason mine would have been.

So the shape you identified has two instances, not one, and the one already on the main line is the more serious.

## What landed

`written` rather than `argued`, because it has to read correctly at both kinds of call site — a comment paragraph and a markdown heading — and "argued" only fits the first. Same body as your sketch:

```sh
written() {  # written <label> <file> <literal> -- the file as written, # and all
```

All three markdown pins route through it: the one that was red, plus two that pass today only because no `#` happens to fall on the matched line. That rerouting restores behaviour rather than silencing it — renaming the heading in `CLAUDE.md` turns the boundary check red again, which is what it is for.

Also: `beside` was printing `ok armed`, copied from the helper it is not. It says `beside` now.

## Both residual gaps, closed (`a4a14e5`)

**The prose duplication can no longer silently return.** You were right that the count discipline was applied to the code duplication and not to the prose fixed in the same commit. Counted in both directions now — the argument is made once in the guard, and not at all in the report:

| mutation | red |
|---|---|
| the guard's argument dropped | `...is made in the guard, once` |
| a second copy re-added to the report | `and the report does not argue it a second time` |
| the report's pointer sentence dropped | `the report says where the argument was moved to` |
| the `CLAUDE.md` heading renamed | `the extracted section is the boundary section` |

That also retires the last `written` call site this branch added — the counts subsume it. `written` stays for the three markdown pins, which is where the merge showed it was needed.

**The pointer sentence** is held by `beside` rather than `written`, on the reasoning that a pointer which has drifted from the derivation is not doing the job a pointer exists for — the standard the pairing line four lines above is already held to.

Named limit, in the code and here: a count finds the argument *as written*, so a reworded second copy is a second copy uncounted — the same limit `dev_read_count` carries for `git branch -r`.

## State

`dev-05` merged in at `5cda5f1`. Suite: **603 checks, all pass** (~3m now, up from seconds — #64–#66 added a lot of fixture work; noting it, not proposing anything). The original #62 mutation still bites after the merge: changing the glob in one file turns equality and that file's pin red.

https://claude.ai/code/session_01NNtmv9SCTyzRZT3E8J13Km

Issue #36, opened 2026-09-06
Spec: Complete the agent boundary: enforce where a PR may be proposed, and state it where a human can audit it


bgunyel:
## Problem Statement

PR #35 introduces a boundary between what an unattended agent may do to this
repository and what is reserved to Bertan. It went through six rounds of
review (head `b625d64` at the time of writing), and a grilling session afterwards found that four of the claims it now
makes are wrong, unstated, or unenforced.

1. **The boundary describes a rule it does not enforce.** The policy is that an
   agent works in a worktree branch and may propose it only into the active dev
   branch; `main` receives a pull request only from Bertan. Nothing checks a
   pull request's base. Verified against the committed hooks, all four of these
   are permitted today: `gh pr create --base main`, `gh pr edit <n> --base main`,
   a REST `POST` to the pull-requests collection with `base=main`, and the
   `createPullRequest` graphql mutation.

   `gh pr create` with no `--base` is worse than any of these: it silently
   defaults to the repository's default branch, which is `main`. That is not an
   evasion, it is the shape of an ordinary mistake — the exact thing these hooks
   exist to stop.

2. **Worktrees fork from the wrong ref.** `.claude/settings.json` sets no
   `worktree.baseRef`, so `EnterWorktree` defaults to `fresh` and branches from
   `origin/<default-branch>` — `main`, not the active dev branch. A worktree
   branch made that way has never contained the active dev branch's work, so a
   base rule enforced on top of it would be enforcing a fiction.

3. **CLAUDE.md states something untrue.** It says the active dev branch "rests
   on those hooks alone and can rest on nothing else". The reasoning given is
   sound — an agent acts as `bgunyel`, so a server-side rule on `dev-*` cannot
   separate the two — but that premise is a configuration choice, not a law. A
   deploy key or a GitHub App installation token would give the agent its own
   actor, and rulesets take bypass actors. As written, the sentence stops the
   next reader from asking the question at all.

4. **Two things are open and unnamed.** Nothing guards `.claude/`: the only
   `Edit|Write` hook covers three `docs/` directories, so the hook files and
   `settings.json` that carry the boundary are freely editable by the agent the
   boundary constrains. And the hooks have no stopping rule — six review rounds
   each found something, two of them introduced by the fix to the previous two,
   and nothing written down says when to stop looking.

Separately, the vocabulary has gone soft. This repository already uses **probe**
for an empirical measurement of an unknown — `scripts/probe_*.py`, "one
[...]

PR #74, opened 2026-09-11
Stop cutting a command on a separator that is inside quotes

### LINKED ISSUE #68: cs_split cuts on a separator inside quotes: ordinary sed and grep commands are refused as pushes and commits
## Parent

#36

## Context

Found by review of `dev-05`, and reproduced against the hooks as they stand on
`dev-05`. It fired twice in a live session against that session's own ordinary
`sed` commands — editing the hooks is exactly the work that trips it.

## What is wrong

`lib/command-scan.sh:288` splits on separators with a plain character class:

```sh
cs_split() {
  sed -e 's/&&/\n/g' -e 's/||/\n/g' -e 's/[;&|()`]/\n/g' \
```

Nothing there knows about quoting. A `|` inside a quoted argument is a fragment
boundary like any other, so `sed`'s `s|…|…|` delimiter and a `grep` alternation
both cut the command into pieces that were never commands.

For `sed -i 's|git push --all origin|X|' f.sh`, `cs_split` yields:

```
sed -i 's
git push --all origin
X
' f.sh
```

Fragment 2 is at the head of its own line, so `cs_git_args push` finds it and
the hook judges a push that does not exist.

## What it costs

Measured, `0` = allow and `2` = block, in a linked worktree and on `main`
respectively:

```
2  no-git-push.sh        sed -i 's|git push --all origin|X|' f.sh
2  no-git-push.sh        sed -i 's|git push -f origin main|X|' f.sh
2  no-git-push.sh        sed -i 's|git push|X|' f.sh
2  no-git-push.sh        grep -rn 'git push --all|git push -f' .claude/
2  no-commit-to-main.sh  sed -i 's|git commit -m x|X|' f.sh
2  no-commit-to-main.sh  grep -n 'git commit|git push origin main' *.sh
```

**Why it is intermittent, which is worth recording.** The fragment must begin
with the command word once leading whitespace is stripped, so a regex anchor in
the pattern saves the command by accident and a leading space does not:

```
0  sed -i 's|^git push --all origin|X|' f.sh
2  sed -i 's| git push --all origin|X|' f.sh
0  sed -i 's|^git commit|X|' f.sh
2  sed -i 's|git commit|X|' f.sh
```

Two commands doing the same job, one refused and one not, on a difference that
has nothing to do with what either would run. That is the part that reads as
arbitrary from inside a session.

## The permitting direction: swept, and not found

The claim attached to the original report was that this over-refuses only, and
that claim had been checked on `gh pr create` and the REST path alone.
`no-git-push.sh` and `no-commit-to-main.sh` have now been swept too. No
permitting case was found in either, and the structural reason is worth writing
down so the next person does not re-derive it:

- Splitting never deletes text, it only inserts boundaries. Every "is X present
  in any fragment" test still sees every character it saw before.
- The only way a cut can hide evidence is by truncating a verb's argument list.
  `no-commit-to-main.sh` refuses `git commit` on `main` whatever the arguments,
  so a truncated list changes nothing. `no-git-push.sh` refuses on flags —
  `--all`, `--mirror`, `-f`, `--delete` — that cannot sit behind a quoted
  free-text argument, because `git push` has no `-m`-style slot to put one in.

**One soft spot, named rather than closed.** `no-git-push.sh:153` records that
"an empty argument list is the permitted case", so a push whose arguments were
lost past a cut would read as a bare push. Probed with `git push "|--all
origin"` and `git push "x|--all" origin`: both still blocked. It holds — but it
holds because of a property of `git push`'s own CLI, not because of anything
this library does, so the honest phrasing is *swept and not found*, never
*cannot happen*. Any future hook that judges a verb by a free-text argument
reopens this question.

## What to build

Teach `cs_split` which separators are inside quotes, and cut only on the ones
that are not. The quote state is readable at that point: `cs_normalise` does not
strip quotes, and the hooks that strip them do so downstream of the split.

Two constraints on the shape of the fix:

- **Fail in the refusing direction.** Any input the tracker cannot read —
  unbalanced quotes above all — must fall back to splitting exactly as today.
  Over-refusing is the behaviour being complained about here, and it is still
  the right answer when the text cannot be read.
- **A closed quote reopens the separator.** `echo "a" | git push --all origin`
  must still split at that `|`. A tracker that treats everything after the
  first quote as quoted converts this ticket into a real hole.

## Not in scope

Writing a shell parser. The stopping rule in these files is that a wrapper's
payload is refused outright rather than parsed, and that stays. This is about
one character class that does not know about quoting, not about reading shell
properly. The wrapper detections run on the raw command text *before* the split
in every hook that has one, so a quote-aware `cs_split` does not weaken them —
that ordering is load-bearing and should be asserted rather than assumed.

## Acceptance criteria

- [ ] `cs_split` does not cut on `;`, `&`, `|`, `(`, `)` or a backtick that lies
      inside single or double quotes
- [ ] Unbalanced or unreadable quoting falls back to the current splitting, and
      a check covers it
- [ ] A closed quote restores separator meaning: `echo "a" | git push --all
      origin` still splits at the `|`, and a check covers it
- [ ] `tok` checks in `check-hooks.sh` pin the fragment list for the `sed` and
      `grep` shapes above, both the firing and the anchored non-firing ones
- [ ] `check_in` checks cover the six over-refusals listed above, now ALLOW,
      against `no-git-push.sh` and `no-commit-to-main.sh`
- [ ] The controls stay BLOCK — `git push --all origin`, `git push -f origin
      main`, `git commit -m x` on `main` — and a check covers each
- [ ] A check asserts that each wrapper detection still reads the raw command
      text rather than the split fragments
- [ ] Mutation-checked: reverting `cs_split` to the plain character class turns
      the new ALLOW checks red, and no fix lands without a check that fails
      without the fix
- [ ] `lib/command-scan.sh`'s own header gains the defect and the sweep above,
      including the `no-git-push.sh:153` soft spot stated as swept-not-found

bgunyel:
Closes #68.

`cs_split` cut on separators with a plain character class that knew nothing
about quoting, so a `|` inside a quoted argument was a fragment boundary like
any other. An ordinary sed substitution written with `|` as its delimiter
yielded a push standing at the head of its own line, and every hook refused it:

```
sed -i 's|git push --all origin|X|' f.sh        was refused as a push
grep -rn 'git push --all|git push -f' .claude/  was refused as a push
sed -i 's|git commit -m x|X|' f.sh              was refused as a commit
```

Six such commands, measured in the ticket, and it fired twice in a live session
against that session's own edits — editing these hooks is exactly the work that
trips it. The part that reads as arbitrary from inside a session is that the
fragment has to *begin* with the command word, so an anchor in the pattern saved
the command by accident and a leading space did not: `s|^git push|` was
permitted and `s|git push|` was not. Both spellings are now pinned side by side.

## The permitting direction: swept, and not found

Splitting never deletes text, it only inserts boundaries, so every "is X present
in any fragment" test still saw every character. The only way a cut can hide
evidence is by truncating a verb's argument list, and neither hook judges a verb
that way: `no-commit-to-main.sh` refuses `git commit` on `main` whatever the
arguments, and `no-git-push.sh` refuses on flags that cannot sit behind a quoted
free-text argument.

`no-git-push.sh:153` — "an empty argument list is the permitted case" — is the
one soft spot. Probed both ways; both still blocked. It holds because of a
property of `git push`'s own CLI, not because of anything this library does, so
the header says **swept and not found** rather than "cannot happen".

## Three ways this fix could itself have become a hole

- A tracker treating everything after the first quote as quoted converts the
  ticket into a real hole. A closed quote reopens the separator, and
  `echo "a" | git push --all origin` still splits at the pipe.
- Unbalanced quoting is text this cannot read, and falls back to splitting
  exactly as the character class did. Quote state is per line, which is what
  leaves the multi-line quoted string CLAUDE.md names still refused.
- **Double quotes do not make text inert.** A command substitution inside them
  runs, so protecting a double-quoted span outright would have hidden every
  command written that way — silently and in the permitting direction, the shape
  of nine of the defects these files have already had. A line carrying one goes
  to the same fallback.

That last is a deliberate deviation from the ticket's first checkbox as
literally written. It is argued in the code, in the file header and in the
commit message, and pinned by a check pair. All six measured over-refusals are
single-quoted, so no listed acceptance case is left unmet.

The fallback is that same walk with quote tracking switched off, not a second
encoding of the separator set — which is the shape this file's header names as
the reason the file exists.

## What flipped, and what did not

`git commit -m "wait; then git push --all origin"` on one line is text again and
is permitted. The two checks carrying that flip moved out of the
accepted-false-positive section, where the heading no longer described them, and
sit beside the multi-line spelling that did *not* flip with them.

## Evidence

616 checks, all passing, 54 of them new. Four mutations:

| mutation | red |
|---|---|
| `cs_split` back to the plain character class | 29, all new, the flips among them |
| the substitution fallback removed | 7 — 6 new, plus the accepted false positive about a base written after a substitution |
| the escaped-quote branch removed | 2, both new — without them that branch had no check that failed without it |
| `no-git-push.sh`'s wrapper rule removed | 6 — 2 new, plus the 4 that already covered it |

That last is the check that the wrapper detections still read the raw command
text rather than the fragments, which quote-aware splitting makes load-bearing:
a single-quoted payload is now one fragment with no command position in it.

Two further verdict changes were found by sweeping a corpus of 57 commands
against both versions in two contexts, rather than by the suite — the delimiter
spelled with double quotes, and a substitution inside single quotes, which runs
nothing. Both now have checks. Across that sweep every change was BLOCK to
ALLOW, and every one on a command that manipulates text.

https://claude.ai/code/session_01DPSwjEyZBXrJ5UETM9kk9x

### DIFF
--- a/.claude/hooks/check-hooks.sh
+++ b/.claude/hooks/check-hooks.sh
@@ -32,6 +32,16 @@
 # the workaround is to drop the redirect -- and was filed anyway, because the
 # question it gets wrong is the one lib/command-scan.sh exists to answer once.
 #
+# The quoted-separator checks come from a fifth review, of dev-05, and are the
+# second group to name a defect in the refusing direction. They are also the
+# first whose cost lands on the work of editing these files: cs_split cut on a
+# separator inside quotes, so an ordinary `sed -i` substitution written with |
+# as its delimiter yielded a push that does not exist, and it fired twice in a
+# live session against that session's own edits. The pairs matter more than the
+# individual verdicts there -- the anchored and unanchored spellings of one
+# substitution are pinned side by side, because it was the anchor that decided
+# the verdict and that is the part that reads as arbitrary.
+#
 # So the number below is not a measure of the boundary. A check suite is
 # evidence about the cases it names and about nothing else, and every case here
 # was named by someone who went looking for one it had missed.
@@ -461,6 +471,101 @@ git push --mirror origin' \
 tok 'control word removed, brace group' \
     'git push --mirror origin' \
     "$(printf '{ git push --mirror origin; }\n' | cs_split)"
+# Issue #68. The separator pass cut with a plain character class that knew
+# nothing about quoting, so a | inside a quoted argument was a fragment
+# boundary like any other and the second fragment of a sed substitution was a
+# push standing at the head of its own line. Each of these is one fragment now.
+#
+# The firing and the non-firing spellings are pinned side by side because the
+# pair is the evidence: the fragment had to BEGIN with the command word, so a ^
+# in the pattern saved the command by accident and a leading space did not.
+# Two commands doing the same job, one refused and one not, on a difference
+# that has nothing to do with what either would run.
+tok 'a sed delimiter is not a separator' \
+    "sed -i 's|git push --all origin|X|' f.sh" \
+    "$(printf "sed -i 's|git push --all origin|X|' f.sh\n" | cs_split)"
+tok 'the anchored spelling is the same one fragment' \
+    "sed -i 's|^git push --all origin|X|' f.sh" \
+    "$(printf "sed -i 's|^git push --all origin|X|' f.sh\n" | cs_split)"
+tok 'and so is the leading-space spelling that used to fire' \
+    "sed -i 's| git push --all origin|X|' f.sh" \
+    "$(printf "sed -i 's| git push --all origin|X|' f.sh\n" | cs_split)"
+tok 'a grep alternation is not a separator' \
+    "grep -rn 'git push --all|git push -f' .claude/" \
+    "$(printf "grep -rn 'git push --all|git push -f' .claude/\n" | cs_split)"
+tok 'the commit spelling of the same shape' \
+    "sed -i 's|git commit -m x|X|' f.sh" \
+    "$(printf "sed -i 's|git commit -m x|X|' f.sh\n" | cs_split)"
+tok 'the anchored commit spelling, likewise one fragment' \
+    "sed -i 's|^git commit -m x|X|' f.sh" \
+    "$(printf "sed -i 's|^git commit -m x|X|' f.sh\n" | cs_split)"
+tok 'the forced-push spelling' \
+    "sed -i 's|git push -f origin main|X|' f.sh" \
+    "$(printf "sed -i 's|git push -f origin main|X|' f.sh\n" | cs_split)"
+tok 'the bare-push spelling' \
+    "sed -i 's|git push|X|' f.sh" \
+    "$(printf "sed -i 's|git push|X|' f.sh\n" | cs_split)"
+tok 'the grep alternation over a commit and a push' \
+    "grep -n 'git commit|git push origin main' *.sh" \
+    "$(printf "grep -n 'git commit|git push origin main' *.sh\n" | cs_split)"
+# Double quotes protect a delimiter too. They protect only the separators,
+# never a substitution -- see the two below.
+tok 'a delimiter written with double quotes' \
+    'sed -i "s|git push --all origin|X|" f.sh' \
+    "$(printf 'sed -i "s|git push --all origin|X|" f.sh\n' | cs_split)"
+# A closed quote restores the separator. A tracker that treated everything
+# after the first quote as quoted would convert issue #68 into a real hole.
+tok 'a closed quote reopens the separator' \
+    'echo "a"
+git push --all origin' \
+    "$(printf 'echo "a" | git push --all origin\n' | cs_split)"
+# The two fallbacks, both of which split exactly as the plain character class
+# did. Unbalanced quoting is text this cannot read.
+tok 'unbalanced quoting falls back to the old splitting' \
+    "echo 'unclosed
+git push --all origin" \
+    "$(printf "echo 'unclosed | git push --all origin\n" | cs_split)"
+# And a double-quoted span is not inert: a command substitution inside one RUNS,
+# so a line carrying one goes to the same fallback rather than being protected.
+# Getting this wrong would have hidden every command written that way, silently
+# and in the permitting direction.
+tok 'a substitution in double quotes still splits out' \
+    'echo "$
+gh pr merge 5
+"' \
+    "$(printf 'echo "$(gh pr merge 5)"\n' | cs_split)"
+tok 'a backticked span in double quotes still splits out' \
+    'echo "
+git push --all origin
+"' \
+    "$(printf 'echo "`git push --all origin`"\n' | cs_split)"
+# Single quotes need no such exception -- bash runs nothing inside them -- and
+# an escaped substitution in double quotes is text, which is why the backslash
+# is read before the substitution is looked for.
+tok 'a substitution inside single quotes is text' \
+    "grep -n 'git push|\$(x)' ." \
+    "$(printf "grep -n 'git push|\$(x)' .\n" | cs_split)"
+tok 'an escaped substitution in double quotes is text' \
+    'git commit -m "release \$(date) notes"' \
+    "$(printf 'git commit -m "release \\$(date) notes"\n' | cs_split)"
+# The backslash is read for that one purpose. An escaped separator outside
+# quotes still cuts, exactly as it did before, which is the refusing direction.
+tok 'an escaped separator outside quotes still cuts' \
+    'echo a \
+git push --all origin' \
+    "$(printf 'echo a \\| git push --all origin\n' | cs_split)"
+# The other arm of that branch, which the check above does not reach: an escaped
+# quote outside quotes must not OPEN one. Without this the sed argument that
+# follows sits inside a double quote that never closes, the line is unbalanced,
+# and the fallback splits it into a bare push -- so this pair is the check that
+# fails without the escape branch. Every other escape shape tried gave the same
+# answer with the branch and without it, because the fallback agrees with the
+# protected split whenever the quoting is simple.
+tok 'an escaped quote outside quotes does not open one' \
+    'sed -e s/\"/Q/ -e '"'"'s|git push|X|'"'"' f.sh' \
+    "$(printf 'sed -e s/\\"/Q/ -e %ss|git push|X|%s f.sh\n' "'" "'" | cs_split)"
+check no-git-push.sh ALLOW 'and the same command is not a push' \
+         'sed -e s/\"/Q/ -e '"'"'s|git push|X|'"'"' f.sh'
 tok 'git args, plain' 'origin main' "$(printf 'git push origin main\n' | cs_git_args push)"
 tok 'git args, global option with a separate value' \
     '--all' "$(printf 'git -C /x push --all\n' | cs_git_args push)"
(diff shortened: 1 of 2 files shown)


bgunyel:
## Review: quote-aware `cs_split`

**Verdict: approve.** The fix is correct, the evidence in the description is accurate rather than decorative, and I could not find a permitting-direction hole.

### Acceptance criteria — 9 met, 1 met in substance

| # | Criterion | |
|---|---|---|
| 1 | No cut on `;` `&` `\|` `(` `)` `` ` `` inside quotes | ✅ verified each, both quote styles, plus `&&`/`\|\|` |
| 2 | Unbalanced quoting falls back, with a check | ✅ |
| 3 | Closed quote reopens the separator, with a check | ✅ |
| 4 | `tok` checks pin sed/grep fragments, firing *and* anchored | ✅ 9 of them, pairs side by side |
| 5 | The six over-refusals now ALLOW | ✅ all six, plus 4 repeated from the main checkout |
| 6 | Controls stay BLOCK | ✅ 4 checks, mutation-confirmed |
| 7 | A check asserts **each** wrapper detection reads raw text | ⚠️ see note 2 |
| 8 | Mutation-checked | ✅ exact |
| 9 | Header gains defect, sweep, and the `:153` soft spot | ✅ and the line reference is accurate |
| 10 | `check-hooks.sh` passes | ✅ 616 ok, exit 0 |

### Evidence reproduced rather than taken on faith

Every number in the mutation table is exactly right: reverting `cs_split` → **29 red** (all new), removing the substitution fallback → **7**, removing the escaped-quote branch → **2**, removing `no-git-push.sh`'s wrapper rule → **6**.

I added a fifth mutation the table does not list — removing the *inside*-double-quote escape branch — and it also goes red (`an escaped substitution in double quotes is text`). So the "no fix lands without a check that fails without it" criterion holds for the branch that was not tabulated.

**The permitting direction, swept independently.** I built a differential harness — old split vs new, all three hooks, in both the linked-worktree and main-checkout contexts. Exactly six verdicts flip, all BLOCK→ALLOW, all on text-manipulation or genuinely inert quoting, and identical in both contexts, so nothing is context-dependent.

Then the sharper test: an oracle that runs each command under a stubbed `git` and asks whether bash *actually* executes a push, across 31 quoting shapes. **Zero holes.** The only two disagreements are the deliberate unbalanced-quote fallback, in the refusing direction. Two flips that look alarming on paper — `echo "a\"; git push --all origin"` and `echo '"$(git push --all origin)"'` — were confirmed against real bash to run no push at all.

Portability holds: the awk walk passes under both mawk (the system default here) and nawk.

### Three notes, none blocking

**1. A fourth consumer of `cs_split` went unswept — and it is the one the structural argument does not cover.**

`no-work-on-stale-branch.sh` also splits, and it is the only hook where fragments *withdraw* a carve-out instead of triggering a refusal. The new header argues the permitting direction is safe because "splitting never deletes text, it only inserts boundaries, so every *is X present in any fragment* test still sees every character". That reasoning holds for the two hooks named, but does not transfer as written to a hook where losing a fragment head *retains* an exception rather than dropping a refusal.

Swept it anyway — 17 shapes, oracle on `cd` / `git checkout` / `git switch`: **no holes**, every disagreement an over-withdrawal. So the behaviour is fine and the header is incomplete, not wrong. One sentence naming the third consumer and why it is still safe would close it.

**2. Criterion 7 says "each" wrapper detection; the `armed`/`unarmed` literal pins cover 3 of 4.**

`no-work-on-stale-branch.sh` has a wrapper detection too. It is covered behaviourally — removing it turns `a commit wrapped in a shell` and `a cherry-pick wrapped in eval` red — and because its payload is single-quoted, that BLOCK *is* evidence about where the rule reads from, which is the same argument made for the other three. Substantively met; two more `armed`/`unarmed` lines would meet it literally.

**3. `&` inside quotes, and a backtick inside single quotes, are correct but unpinned.**

The `tok` checks exercise `|`, `;`, `(` and `)`. Both behaviours are right; neither has a check. Cheap to add.

### Out of scope, but worth its own ticket

The wrapper regex anchors at line start or after a separator, so these are all **ALLOW**:

```
sudo sh -c 'git push --all origin'
timeout 5 bash -c 'git push --all origin'
xargs sh -c 'git push --all origin'
env FOO=1 sh -c 'git push --all origin'
nohup sh -c 'git push --all origin'
find . -exec sh -c 'git push --all origin' \;
```

Confirmed unchanged by this PR — identical verdicts before and after — so they are pre-existing, and #68 explicitly scopes wrappers out. Noting them only so that closing #68 is not read as having touched them.

bgunyel:
All three notes are addressed in `adc5598`. One of them turned up a correction to my own first attempt at it.

## 1. The fourth consumer — you were right, and the obvious shape was not the shape

The header argued the permitting direction from "a cut can only refuse more", and that argument holds only where a fragment *triggers* a refusal. `no-work-on-stale-branch.sh` withdraws a carve-out from its fragments, so a lost fragment head **retains** an exception. The header now names it, and says which of the two arguments is load-bearing where: it is the fallback, not the cut rule, that carries this hook — a `cd` bash would actually run sits either outside quotes, where it is still cut, or inside a substitution, which sends the whole line to the fallback.

I went looking for the changed verdict and guessed wrong first: a quoted `cd` in the merge's own message. That shape is unreachable — `-m` withdraws the carve-out by itself, so the command is refused before and after, for an unrelated reason. I had already written it as a check when the suite told me it was still BLOCK.

Measured on the suite's own stale fixture, old split against new, the shape that does reach it is a quoted separator in a **sibling** command:

```
git merge origin/dev-05 && echo 'x; cd /tmp'             BLOCK -> ALLOW
git merge origin/dev-05 && echo 'x; git checkout main'   BLOCK -> ALLOW
cd /tmp && git merge origin/dev-05                       BLOCK, unchanged
git checkout main && git merge origin/dev-05             BLOCK, unchanged
```

bash runs no `cd` in either of the first two. All four are checks; the unquoted pair is what says the withdrawal itself still works.

## 2. Criterion 7, in letter

`no-work-on-stale-branch.sh` has the `armed`/`unarmed` pair the other three have. A fifth mutation joins the table: removing its wrapper rule reddens 3 — the new literal, plus `a commit wrapped in a shell` and `a cherry-pick wrapped in eval`.

## 3. `&` and a backtick inside single quotes

Both pinned, as a fragment list and as a verdict. Both have a real spelling, which is why they were worth more than a pin: `&` is a legal sed delimiter, and `grep -rn '`git push --all origin`' docs/` is how you would search the docs for one.

```
sed -i 's&git push --all origin&X&' f.sh    was refused as a push
grep -rn '`git push --all origin`' docs/    was refused as a push
```

## On the out-of-scope finding

Confirmed independently: those six are unchanged by this PR, and the wrapper regex anchoring is pre-existing. Happy to open a ticket for it — say the word and I will, rather than filing one against your review unasked.

626 checks, all passing.

https://claude.ai/code/session_01DPSwjEyZBXrJ5UETM9kk9x

bgunyel:
Verified `adc5598` independently: 626 checks pass, the fifth mutation reddens exactly 3, and both new `(was BLOCK)` labels hold under direct probe rather than only by label. I also checked the correction you made to your own first attempt — `git merge origin/dev-05` is ALLOW but the same merge with any `-m` is BLOCK regardless of content, so the quoted-`cd`-in-the-merge-message shape really is unreachable and the sibling-command shape really is the one that reaches it. Re-ran my own oracles against the new commit: the 31-shape push oracle is unchanged at 29 ok / 2 deliberate over-refusals, and the carve oracle shows no holes. All three notes are closed. Approving.

## Yes — please open the ticket. Scope it as follows.

The reason it is worth filing is stronger than "six commands are permitted". **`cs_split` already maintains the list**, in the stage-2 prefix strip:

```
env|command|xargs|nohup|nice|time|stdbuf|ionice|sudo|doas|setsid|chronic    (+ timeout|flock)
```

The four wrapper regexes do not consult it. They answer "is this a wrapper that runs another command?" independently and more narrowly — `(ba|z|)sh -c`, `sh <<`, or `eval`, anchored at `^` or after a separator. That is the same question answered in two places with two different answers, which is the defect class this file's header opens by naming. The asymmetry it produces:

```
BLOCK   sudo git push --all origin           cs_split strips sudo, the push is at a command position
ALLOW   sudo sh -c 'git push --all origin'   cs_split strips sudo, leaves sh -c '...', and the
                                             wrapper rule never fires because sh is not at ^
```

`sudo` is recognised in one half of the library and invisible in the other. Same for `timeout`, `xargs`, `nohup`, `env`. And it is one layer out from the header's own PR #35 defect #2 — *"wrapper re-admission sufficed for a heredoc and did nothing for `sh -c`"* — which now handles `sh -c` and does nothing for `sudo sh -c`.

### What to build

Extend the anchor to admit the prefix words the library already knows, **reusing that list rather than restating it**. A second copy of those twelve words in four hook regexes would be the same defect one more time.

### Two constraints, both load-bearing

**Do not simply drop the anchor.** That reintroduces #68: `grep -rn "sh -c" .claude/` and `sed -i 's|sh -c|X|' f.sh` would be refused as wrapped commands — the same over-refusal class, landing again on the work of editing these hooks. The anchor must widen to a known prefix set, not disappear. A check should pin that these stay ALLOW.

**The list cannot be complete, and the ticket should say so rather than imply otherwise.** These are all ALLOW and will stay ALLOW:

```
python3 -c 'import os; os.system("git push --all origin")'
perl -e '...'                 a make target        a script written to a file and then run
```

The stopping rule is the one already in these files: a wrapper's payload is refused rather than parsed, and the wrapper words are the ones the library can already name. Everything past that is out of reach, and the honest phrasing is *named and not closed*, in the manner of the `no-git-push.sh:153` soft spot — not a claim the set is exhaustive.

### Not in scope

Making these hooks proof against a command written to evade them. They guard against a careless command, not a determined one; anyone with Bash has a dozen exits and the file should not pretend otherwise.

### Priority

Backlog, not next. The realistic trigger is `timeout 30 bash -c '...'` around a network command or `xargs sh -c` in a loop — occasional, not routine — and there is work ahead of it that is not hook work.

### Suggested acceptance criteria

- [ ] The wrapper detections admit a prefix from the set `cs_split` already strips, without a second copy of that set
- [ ] `sudo sh -c`, `timeout 5 bash -c`, `xargs sh -c`, `env sh -c`, `nohup sh -c` each BLOCK a wrapped push, and a check covers each
- [ ] The unwrapped controls still BLOCK, and the #68 shapes still ALLOW — `grep -rn "sh -c" .claude/` above all
- [ ] The out-of-reach shapes are named in the header as such, with `python3 -c` given as the example
- [ ] Mutation-checked: reverting the anchor turns the new BLOCK checks red, and dropping it outright turns the #68 ALLOW checks red

Reference #68 and the header's PR #35 defect #2 in the context section; parent is whatever #68 hangs from.

bgunyel:
Filed as #79, parented to #36, labelled `bug` + `ready-for-agent`, with your scope, constraints and acceptance criteria carried over. Two places where measuring it changed what the ticket says, both worth flagging rather than quietly fixing.

## Your constraint is right; the example given does not demonstrate it

`grep -rn "sh -c" .claude/` stays **ALLOW** with the anchor dropped. The wrapper rule is a conjunction — wrapper pattern *and* a `git … push` pattern in the same raw text — and that command names no push, so the anchor is not what is protecting it.

The shapes that do regress are the ones naming a wrapper word and a push together, which is what a session working on these hooks actually writes. Measured, anchor present vs dropped:

```
grep -rn 'sh -c .*git push' .claude/hooks/                        ALLOW -> BLOCK
grep -rn 'eval .*git push' .claude/hooks/                         ALLOW -> BLOCK
echo 'the eval rule refuses git push --all origin' >> notes.md    ALLOW -> BLOCK
```

Those are what the ticket asks to be pinned.

## Found while scoping it: the anchor is quote-blind, so #68 survives inside the wrapper rule

The wrapper rule reads raw text by design — I pinned that ordering this week — but its anchor carries its own separator class `[;&|(`]`, and that class knows nothing about quoting. The verdict still turns on a sed delimiter:

```
BLOCK   sed -i 's|sh -c git push --all|X|' f.sh     the | before sh -c satisfies the anchor
ALLOW   sed -i 's/sh -c git push --all/X/' f.sh     same command, different delimiter
```

Same complaint as #68, one rule over. `cs_split` could not fix it, because this is the one rule that must not use `cs_split`. It is in #79 as a section of its own and as an acceptance criterion — *decide* it, either by fixing it or by naming it in the header as a soft spot the way `no-git-push.sh:153` is — rather than folded into your scope, since it is a different question about the same expression.

## Also reproduced

`env FOO=1 sh -c` is ALLOW but bare `FOO=1 sh -c` is already BLOCK — the anchor admits environment assignments, so it is the wrapper *word* that is missing, not the assignment. The ticket says so, to save the next person the probe.

All six of your ALLOW shapes reproduce exactly, as does the `sudo git push` / `sudo sh -c` asymmetry that makes the case.

https://claude.ai/code/session_01DPSwjEyZBXrJ5UETM9kk9x

PR #77, opened 2026-09-11
Read the merge settings instead of recording them by hand

### LINKED ISSUE #71: Check the merge settings instead of recording them by hand: the guard's header is now false in the opposite direction
## Parent

#36

## Context

Found by examining the guard's header against the live repository settings.
The same paragraph was corrected once already, two days ago, for the same
defect in the opposite direction.

## What is wrong

`.claude/hooks/no-work-on-stale-branch.sh:35-41` states:

> That is a repository setting and not something this file can assert. At the
> time of writing it is NOT yet applied -- both are still enabled -- and issue
> #44 carries it as an acceptance criterion for Bertan, because changing a
> repository setting is not an act an agent takes. Until it is applied, a branch
> merged by squash or by rebase defeats this detector [...]

`gh api repos/bgunyel/clause-and-effect` returns `allow_squash_merge: false`
and `allow_rebase_merge: false`. Both are applied.

Three claims in that paragraph are now false, not one:

| claim | actual |
|---|---|
| "both are still enabled" | both disabled |
| "#44 carries it as an acceptance criterion" | #44 closed `2026-09-09T18:06:11Z` |
| "Until it is applied, a branch merged by squash or by rebase defeats this detector" | it is applied; the fallback detector is sound by configuration |

**Both durable records are wrong, not just the header.** #36's Amendment
designates exactly two:

> A bash hook cannot assert a GitHub setting without a network call, so these
> have no check behind them and this map plus the guard's header are their only
> durable record.

The map's own entry reads *"`allow_squash_merge` and `allow_rebase_merge` are
to be **disabled** by #44"* — future tense — while the line directly above it
reads *"`delete_branch_on_merge` is **enabled** (done)"*. The map has a
done-marker convention that this line never received. So neither designated
record states the current truth, and one asserts the opposite of it.

## Why it matters

No runtime impact, and the error points the safe way: a reader under-trusts a
detector that works. `check-hooks.sh`'s `armed()` strips comments before
grepping (`sed 's/[[:space:]]*#.*$//'`, line 209), deliberately, so no check
reads this text and nothing in the hook's executable path depends on it.

What matters is that this is the **second firing of the same defect in the same
paragraph within 48 hours**. `d840e57` — *"stop claiming a setting that is not
set"* — exists because review caught the original wording flatly asserting
*"allow_squash_merge and allow_rebase_merge are disabled on the repository"*
when they were not. The fix applied was to correct the sentence. That left the
mechanism untouched: a hand-maintained prose record of live remote state, with
nothing able to detect drift. It drifted again, the other way, and three
commits to this file on 2026-09-11 (`c7accb8`, `b5090df`, `4eebeab`) passed
over the paragraph without revising it.

The header convicts itself:

> An assumption a repository setting can silently break is not an assumption;
> it is a bug with a delay.

A *record* a repository setting can silently falsify is the same bug with the
same delay. The chosen mitigation reproduces the disease it was prescribed for.
The next drift is the one that matters — if someone re-enables squash merging,
this paragraph will say nothing, because nobody updates it.

The Amendment states its own purpose as *"so that the map does not read as
current while being wrong."* Both halves of its record now do exactly that,
which is why this is filed against #36 rather than as a loose correction.

## Suggested fix

Correcting the two texts is necessary and not sufficient; alone it schedules
the third firing.

**1. Read the settings where a network call is allowed.**
`report-stale-branches.sh` already runs at `SessionStart` and already opens with
a pruning fetch. It is the one component in this system permitted a network
call, and therefore the only place a GitHub setting can be checked rather than
recorded. Have it read `allow_squash_merge`, `allow_rebase_merge` and
`delete_branch_on_merge` and report any that has drifted from the value the
lifecycle rule requires. This converts the claim into the form this repository
already prefers:

> the keying claim stated as a check rather than as a sentence in a header
> — `check-hooks.sh:1507`

It must fail open on the same timeout as the fetch: an offline session starts,
and its report says the settings could not be read. #44 already established that
this script's arming properties are asserted as literals in the check suite
rather than guarded by a hook on `.claude/`; the same treatment applies to the
new read.

**2. Correct the header** to state the settings as applied, drop the
"#44 carries it" pointer now that #44 is closed, and drop the "Until it is
applied" hazard paragraph. If step 1 lands, the header should point at the
drift report as the live answer rather than restating a value it cannot verify.

**3. Correct the map** with a dated append to #36, following the Amendment's own
convention — it is itself a dated append, and it leaves the Solution section
above it as written. Marking the entry `(done)` in place would edit a record
rather than extend it.

## Acceptance criteria

- [ ] `report-stale-branches.sh` reads the three merge-related repository
      settings and reports any that has drifted from what the lifecycle rule
      requires
- [ ] The settings read fails open on timeout, and the report says so
- [ ] A check asserts the new read as a literal, in the same manner as the
      existing arming checks
- [ ] Mutation-checked: removing the read turns its check red
- [ ] `no-work-on-stale-branch.sh`'s header no longer claims either setting is
      unapplied, and no longer points at #44 as open
- [ ] #36 carries a dated append recording that both settings are applied
- [ ] The check suite passes

## Evidence

- `.claude/hooks/no-work-on-stale-branch.sh:35-41` — the three false claims
- `gh api repos/bgunyel/clause-and-effect --jq '{squash: .allow_squash_merge, rebase: .allow_rebase_merge}'` → `{"squash": false, "rebase": false}`
- #36, *Amendment — 2026-09-09* — "this map plus the guard's header are their only durable record"; the entry still reads "are to be **disabled** by #44"
- `d840e57` (2026-09-09) — *"stop claiming a setting that is not set"*, the first firing, in the opposite direction
- #44 — closed `2026-09-09T18:06:11Z`
- `.claude/hooks/check-hooks.sh:209` — `armed()` strips comments, so no check reads the header
- `.claude/hooks/check-hooks.sh:1507` — "stated as a check rather than as a sentence in a header"

Note: #44's unticked acceptance-criteria checkbox is *not* evidence here — no
issue in this tracker ticks one (#43: 0/8, #44: 0/17). The stale text lives on
`dev-05` only; the file does not exist on `main`.

bgunyel:
Closes #71.

#36's Amendment recorded three repository settings in prose because "a bash
hook cannot assert a GitHub setting without a network call". That is true of a
`PreToolUse` hook, which has five seconds. It was never true of
`report-stale-branches.sh`, which runs at SessionStart and already opens with a
pruning fetch.

The recorded version failed twice inside 48 hours, once in each direction —
`d840e57` corrected a flat claim that both merge settings were disabled when
they were not; #71 found that correction false once they were, after three
commits to the file in one day passed over the paragraph. Correcting the prose a
third time schedules the third firing, so the claim becomes a check.

## What changed

- **`report-stale-branches.sh` reads the three settings** each session and names
  any that has drifted, with the detector that rests on it. Bounded by its own
  10s budget; fails open exactly as the fetch does. Verified against `gh`
  absent, `gh` failing, `gh` hanging (killed at 10s), and a read returning empty
  or `null` fields.
- **The guard's header states no value.** It points at the report. Its
  consolation is also narrowed to the case it covers: *"the other detector still
  catches it"* is false when the setting that drifted is
  `delete_branch_on_merge`, which removes the gone detector itself.
- **The history is told once and counted** — one copy in the guard, zero in the
  report, which carries a pointer. Three retellings in three wordings had
  already diverged, which is the defect this ticket is about.
- **The hook timeout goes 30 → 40.** Two serial network calls took the margin
  over the budgets from 15s to 5s. A cap is not a wait; the hook still exits
  when done.

## Evidence

603 → 619 checks, all passing (17 check invocations added, 1 removed). Fifteen mutations, each turning its own check
red and nothing else: the read deleted and commented out, a setting dropped, a
required value flipped, the bound dropped, a budget raised and renamed, the
message softened, the hook timeout lowered, the guard's pointer replaced by a
recorded value, the history deleted, retold, and its pointer removed, the two
headings folded back together, and `MISMATCHED` set on the unread arm.

**Review follow-up (8856463).** An unread field printed under the `DRIFTED`
heading, telling a skimmer a setting had changed when the truth was that it was
unknown. `drift` now sets `MISMATCHED` only where a value came back wrong, and
an unread field reports under its own heading. Worth noting how that fix was
checked: two `armed` literals on the new heading both stayed green when
`MISMATCHED` was also set on the unread arm — the regression that puts an
unknown back under `DRIFTED` — because `armed` cannot ask what *else* is on a
line, and that mutation adds rather than removes. The arm is therefore extracted
and compared as a string, the way the dev derivation is; the mutation went 0 red
before that pin and 1 red after.

Two limits are named rather than closed. The six value pins read `drift`'s call
sites, not its body — the only seam that would drive the function is sourcing
the report, which runs the fetch. And commenting out the read leaves the jq pin
green, because `armed` strips from a `#` on the line it reads and the filter
sits on a continuation line; that measurement is recorded in the suite, and it
is why the read and its bound are pinned separately.

`make test`: 574 pass, 1 pre-existing failure (`test_installed_packages_match_uv_lock`,
local venv drift — no Python or lockfile is touched here).

https://claude.ai/code/session_01DRpSMo37qJPB9opP6qvBzf

### DIFF
--- a/.claude/hooks/check-hooks.sh
+++ b/.claude/hooks/check-hooks.sh
@@ -306,6 +306,11 @@ beside() {  # beside <label> <file> <literal>
   fi
 }
 
+# A guard for the one check below that compares two numbers rather than
+# matching a literal. An absent derivation must not reach `[ -gt ]`, which
+# errors rather than answering.
+numeric() { case "$1" in ''|*[!0-9]*) return 1 ;; *) return 0 ;; esac; }
+
 tok() {  # tok <label> <expected> <actual>
   if [ "$3" = "$2" ]; then
     printf '  ok   %s\n' "$1"
@@ -2047,6 +2052,71 @@ armed 'the fetch is bounded, so an offline session still starts' \
   "$HOOKS/report-stale-branches.sh" 'timeout "$FETCH_TIMEOUT" git fetch'
 armed 'a failed fetch says so, because neither detector is armed after one' \
   "$HOOKS/report-stale-branches.sh" 'FAILED or timed out'
+
+# THE MERGE SETTINGS, READ RATHER THAN RECORDED. Both detectors rest on a
+# repository setting no bash hook can assert, so #36's Amendment recorded the
+# three values in prose -- "these have no check behind them and this map plus
+# the guard's header are their only durable record." Why that was the wrong
+# answer is argued in no-work-on-stale-branch.sh's header and is not retold
+# here; the pins below hold it where it now lives.
+#
+# Be exact about what these are: pins on the report's code, in the same manner
+# as the fetch pins above. They are not evidence about the repository's
+# settings and cannot be. Nothing in `.claude/` can be that evidence, and a
+# file claiming to be is the defect these replace.
+armed 'the report reads the merge settings rather than trusting a record of them' \
+  "$HOOKS/report-stale-branches.sh" "gh api 'repos/{owner}/{repo}'"
+armed 'and reads all three the branch lifecycle rule depends on' \
+  "$HOOKS/report-stale-branches.sh" \
+  '[.allow_squash_merge, .allow_rebase_merge, .delete_branch_on_merge]'
+armed 'the settings read is bounded, so an unreachable API still starts the session' \
+  "$HOOKS/report-stale-branches.sh" 'timeout "$SETTINGS_TIMEOUT" gh api'
+armed 'a settings read that did not happen says so, rather than reading as fine' \
+  "$HOOKS/report-stale-branches.sh" 'merge settings: NOT READ'
+# What the pins above are worth, measured rather than reasoned: commenting out
+# the read turns the first and the third red and leaves the second -- the
+# settings list -- green, because `armed` strips from a `#` on the line it is
+# reading and the jq filter sits on a continuation line that no one commented.
+# That is why the read and its bound are pinned on their own line rather than
+# the settings list being trusted to stand for all three.
+#
+# The required values, one line each, because the read alone says nothing about
+# what it is compared against -- and a fixed string spanning two lines is
+# satisfied by a file holding either one, measured on a two-line fixture for the
+# derivation pins below.
+armed 'squash merging must be off, or the fallback detector is unsound' \
+  "$HOOKS/report-stale-branches.sh" 'drift allow_squash_merge "$SQUASH" false'
+armed 'rebase merging must be off, for the same reason' \
+  "$HOOKS/report-stale-branches.sh" 'drift allow_rebase_merge "$REBASE" false'
+armed 'and delete_branch_on_merge must be on, which is what the gone detector reads' \
+  "$HOOKS/report-stale-branches.sh" 'drift delete_branch_on_merge "$DELETE" true'
+# The guard's end of it. Its header carried the value in prose and drifted
+# twice; what replaces that is a pointer to the read above, and a pointer is the
+# thing a later edit deletes on its way to writing a value back down. `written`
+# rather than `armed`: this is prose in a comment, which is the whole of what it
+# asserts, and stripping comments would erase the line rather than a remark.
+written 'and the guard points at that report instead of recording a value itself' \
+  "$HOOKS/no-work-on-stale-branch.sh" 'That report is the live answer, and'
+# And the same counting the dev-branch argument gets below, for the same reason
+# and against the same failure. This history is the one thing in the change that
+# is prose rather than code, which is what the last two records of it were --
+# a third copy would be the defect the ticket is about, arriving inside its own
+# fix. It is told in the guard, counted at one there and at zero in the report,
+# and the report carries the pointer instead. This suite tells it nowhere, so
+# there is nothing here to count: a literal written out below would count
+# itself.
+HISTORY='it asserted both settings disabled before they were'
+tok 'the history of the recorded version is told in the guard, once' \
+    '1' "$(prose_count "$HOOKS/no-work-on-stale-branch.sh" "$HISTORY")"
+tok 'and the report does not tell it a second time' \
+    '0' "$(prose_count "$HOOKS/report-stale-branches.sh" "$HISTORY")"
+written 'the report says where that argument lives instead' \
+  "$HOOKS/report-stale-branches.sh" "is in no-work-on-stale-branch.sh's header"
+# What these six pins are NOT evidence of, named because the suite is evidence
+# about the cases it names and nothing else: they read the call sites, not
+# `drift` itself. Mutate that function to return early and all six stay green.
+# Pinning its body was considered and rejected -- the only seam that would drive
+# it is sourcing the report, and sourcing the report runs the fetch.
 # Read-only by name and by content. The name is checked by being the path above;
 # the content is checked here.
 unarmed 'the report removes no worktree' \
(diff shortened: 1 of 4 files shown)


bgunyel:
## Review: all seven acceptance criteria met

Verified against #71 by running the suite on the branch, running twelve mutations of my own plus the two limits this PR declares, and exercising seven runtime paths of the new read. Testing was done in throwaway clones.

| # | Criterion | |
|---|---|---|
| 1 | report reads the three settings and names drift | ✅ `gh api 'repos/{owner}/{repo}'`, all three fields, each drift named with the detector resting on it |
| 2 | fails open on timeout, and says so | ✅ three ways — `gh` absent → *"no gh on PATH"*; `gh` exit 1 → *"failed or timed out"*; `gh` hanging → killed at exactly 10s. Exit 0 in every case |
| 3 | a check asserts the read as a literal | ✅ thirteen net new checks, in the `armed`/`written`/`tok` idiom #44 established for this file |
| 4 | mutation-checked: removing the read turns its check red | ✅ and further, below |
| 5 | guard header drops the unapplied claim and the #44 pointer | ✅ no surviving match for `#44`, "not yet applied", or "still enabled" |
| 6 | #36 carries a dated append | ✅ *Amendment — 2026-09-11*, leaving the prior Amendment as written, per its own convention |
| 7 | the check suite passes | ✅ 616 ok, `ALL CHECKS PASSED` |

### Mutations

Twelve, each turning its own check red and nothing else: deleting the read (3 red); dropping a setting from the jq list (1); flipping a required value (1); dropping the `timeout` bound (1); renaming a budget (3 — the `END` guard holds, a lost term prints nothing rather than a smaller sum); raising `FETCH_TIMEOUT` past the hook cap (2); softening the NOT READ message (1); lowering the hook timeout (1); deleting the history, retelling it in the report, and removing either pointer (1 each).

**Both declared limits reproduce exactly as described.** Commenting out the read does leave the settings-list pin green while the read pin and its bound go red — the filter sits on a continuation line `armed` never reaches. Mutating `drift` to return early does leave all six value pins green. Declared rather than hidden, which is the right treatment for both.

### Runtime

A `null` field, a read that exited 0 with nothing behind it, real drift, and the clean state each produce the intended sentence. Live settings are `squash:false, rebase:false, delete_on_merge:true`, so the report prints *"as required"* on the next session start.

### Two corrections, neither blocking

**1. The count in the description is wrong.** It reads "613 → 616 checks". This branch is a clean fast-forward of `dev-05` — the merge base *is* `dev-05`'s tip — and `dev-05` measures **603**. The change adds 14 check invocations and removes 1, so it is **603 → 616**: ten more than claimed. The end state is right and the direction is right; the baseline is not. Worth fixing here specifically, because a hand-written number that has drifted from the thing it describes is the defect this ticket exists to close, and this one appears in the PR that closes it.

**2. An unread field prints under a `DRIFTED` heading.** When the API answers but returns `null`, the per-setting line correctly reads *"was not reported by the API"* — but it sits beneath *"merge settings: DRIFTED from what the branch lifecycle rule requires"*. A reader who skims the heading concludes a setting changed when the truth is that it is unknown. The direction is safe (alarm, not false comfort) and the line itself is explicit, so this is a wording nit rather than a fault in the design.

### One judgment call, recorded rather than objected to

Worst-case session-start latency against a dead network goes from 15s to 25s of blocking, capped at 40s. The argument in `check-hooks.sh` is correct — a cap is not a wait, and the hook exits when done — but the floor on a genuinely unreachable network did rise by ten seconds. That is the price of the fix, and it looks worth paying.

### On the class, not just the two records

Every remaining mention of the three settings across `CLAUDE.md`, `.claude/skills/branch-hygiene/SKILL.md` and the hooks states a requirement or a mechanism — *"the fallback REQUIRES allow_squash_merge and allow_rebase_merge to be…"* — and never a live value. No third durable record survives to drift next. The change closes the class rather than the two records #36 designated, which is what the ticket asked for and slightly more than it listed.

bgunyel:
Both corrections taken. Pushed as 8856463.

**1. The count.** You are right, and I verified it rather than taking the
number: diffing `check-hooks.sh` against the merge base showed 14 check
invocations added and 1 removed, so 603 + 13 = 616. The 613 was an intermediate
state — the count after the first implementation pass and before the three
checks the pre-merge review added — which I then wrote into the description as
if it were the baseline. The commit message had 603 right; only the description
was wrong.

Noted where it stings: a hand-written number that had drifted from the thing it
describes, in the pull request that closes a ticket about hand-written records
drifting. Description corrected, and it is now **603 → 619** after the second
fix below.

**2. The unread field under a DRIFTED heading.** Fixed rather than left as a
nit, because it is this ticket's own defect one level up from the sentence that
was careful about it. `drift` sets `MISMATCHED` only where a value came back and
was the wrong one; an unread field now reports under its own heading:

```
merge settings: NOT FULLY READ -- the API answered without these, so what
       rests on them is unknown rather than wrong:
  allow_squash_merge was not reported by the API; the rule requires false -- ...
```

A real mismatch beside an unread field still reaches `DRIFTED`, which is right —
something did change. All four paths re-checked: three fields unread, mixed,
drift alone, clean.

**The part worth reporting, because the suite failed at it first.** I pinned the
fix with two `armed` literals — the new heading, and the `MISMATCHED` test. Both
are green against folding the headings back together. Both stayed **green** when
I set `MISMATCHED` on the unread arm as well, which is precisely the regression
that puts an unknown back under `DRIFTED`. `armed` asks whether a literal is
somewhere in a file and cannot ask what *else* is on the line, and the mutation
that matters here adds rather than removes.

So the arm is extracted and compared as a string, the way the dev derivation is
— the only shape of pin here that sees an addition. Measured: that mutation went
0 red before the new pin and 1 red after. Two pins that agreed with each other
and with nothing else is the shape this suite exists to not have, and it was in
the fix for your own finding.

**On the latency call** — recorded and not argued with: the floor against a dead
network did rise 15s → 25s. That is the price, and the two `timeout` calls are
what bound it; the 40 is only a cap.

Suite at 619, `ALL CHECKS PASSED`. Fifteen mutations now, each red for its own
reason.

https://claude.ai/code/session_01DRpSMo37qJPB9opP6qvBzf

PR #78, opened 2026-09-11
Land the Amendment's stage-3 re-scope: the branch lifetime, and the sweep the hooks cite

### LINKED ISSUE #70: Land the Amendment's stage-3 re-scope: CONTEXT.md omits the branch lifetime, and nothing holds Bertan's sweep
## Parent

#36

## Context

Found by reading #36's Amendment against the tree it produced, 2026-09-11.

## What is wrong

#36's Amendment re-scoped stage 3 in one sentence:

> **Stage 3's scope changes.** #41 absorbs the branch-lifecycle vocabulary and
> the worktree sweep alongside the rotation rewrite and the standard of care it
> already owned, so that one file has one owner and the code/documents split of
> this specification survives.

Neither half arrived. #41's acceptance criteria were never rewritten to carry
them, and #41 shipped before #44 existed, so its implementer had nothing to
document even had the criteria asked.

| when (UTC) | what |
|---|---|
| 09-09 11:36:52 | #44 filed |
| 09-09 11:37:32 | #41 body edited — a one-line repair to a broken `Blocked by` reference, nothing else |
| 09-09 11:38:44 | the Amendment lands on #36, assigning #41 the lifecycle vocabulary and the sweep |
| 09-09 14:46:35 | **PR #48 merges #41** — `CONTEXT.md` +34, `CLAUDE.md` +12, `branch-hygiene` rewritten |
| 09-09 18:05:53 | **PR #57 merges #44** — three hooks, `settings.json`, `branch-hygiene` +13; no `CONTEXT.md`, no `CLAUDE.md` |

#41's body still carries the eight criteria it was filed with on 09-06. The
edit seventy-two seconds before the Amendment is the tell: the body was open,
and the new scope did not go into it.

The Amendment had a safeguard against exactly this and it was not followed. It
set the revised order **#38 → #43 → #44 → #39 → #40 → #41**, #41 last so that
it could document what the earlier tickets built. Execution ran #38 → #39 →
**#41** → #43 → #47 → #50 → #51 → #40 → #44. #41 went third.

### 1. `CONTEXT.md` does not say when a worktree branch ends

The *Worktree branch* entry defines what the branch is and why the permission
keys on the linked worktree rather than the name. It stops there. The lifetime
— one pull request, then the branch is finished — is absent, as is the rule
against reusing a worktree that produced it.

The rule itself is written down twice, in neither place a reader looking for
vocabulary would go:

- `.claude/hooks/no-work-on-stale-branch.sh:2` — "A worktree branch exists for
  one pull request. When that pull request merges, the branch has served its
  purpose..."
- `CLAUDE.md:239` — "A worktree branch lives as long as its pull request, and
  work moves to a new one once that has merged."

The `CLAUDE.md` sentence is **two days old** and arrived sideways: `git log -S`
finds it entering in `765a637`, on PR #66 for #63 — a ticket about naming the
four boundary hooks. Before that, `CLAUDE.md` carried no lifecycle statement at
all. So the rule reached the documents by accident of an adjacent fix, not
because anything owned it, and the glossary the other two files point at is
still the one place it is missing.

That pointer is the cost. `branch-hygiene` says "`CONTEXT.md` defines all three
terms" and `no-work-on-stale-branch.sh` sends a reader to "CONTEXT.md's
*worktree branch* entry". Both are true about the keying and silent about the
lifetime, so a reader who follows the pointer learns everything about a worktree
branch except when it dies — and the hook that refuses their commit is enforcing
a rule the glossary does not contain.

### 2. "Bertan's sweep" is cited by two hooks and exists nowhere

This is the half that did not merely go undocumented — it went unwritten, and
shipped code points at it.

- `no-work-on-stale-branch.sh:3-4` — a merged worktree branch is removed
  "remotely `delete_branch_on_merge` removes it, locally **Bertan's sweep**
  does."
- `report-stale-branches.sh:2` — "The read-only half of the branch-hygiene
  **sweep**", and at :21 it explains that "The name is `report-` and never
  `sweep-`" because it removes nothing.

There is no other half. `branch-hygiene` has no worktree-sweep procedure:
`git worktree remove` appears nowhere in the repository's documents, and
`git branch -d` appears once, in the rotation, for `dev-NN`. Bertan's procedure
runs confirm-merge → main → new branch → delete the merged `dev-NN` → verify,
and its closing invariant merely asserts that "every other branch present is a
worktree branch" without saying who removes the merged ones or when.

So the local half of the branch lifecycle is load-bearing in a guard's stated
design and unperformed in practice. Merged worktree branches accumulate locally
until someone notices; `report-stale-branches.sh` will name them every session,
correctly, and no documented procedure acts on the report.

### 3. Two shipped files cite a reserved act that `CONTEXT.md` does not list

The same gap has a third, sharper edge. `CONTEXT.md`'s *reserved act* entry
enumerates four: advancing the active dev branch on the remote, merging any pull
request, rotating the dev branch, and publishing a release. `branch-hygiene`
treats that enumeration as closed — "Rotating the dev branch is one of the
**four acts** `CONTEXT.md` names".

Removing a worktree is not one of the four. Two files say it is:

- `report-stale-branches.sh:19-20` — "Removing a worktree or a branch is
  Bertan's -- a reserved act in CONTEXT.md"
- `report-stale-branches.sh:159` — printed into every session's transcript:
  "Nothing was removed. Deleting a branch or a worktree is a reserved act --"

The entry's closing lines cover the neighbouring case and stop short of this
one: they name "the local half of a rotation, `git branch -d`" as reserved
though nothing refuses it. A worktree branch is deleted by the same command, and
removing its worktree by a command the entry never mentions. Under the closed
reading the citation is simply wrong; under the generous reading the entry is
the one place a reader checks and it does not answer. Either way a `SessionStart`
hook is telling the reader to consult a list that does not contain the item.

Adding the fifth act is a one-line fix and it is the same missing scope: what
happens to a worktree branch at the end of its life.

## Why it matters

The vocabulary half is small and the consequence is bounded: enforcement works,
and an agent that reuses a merged worktree branch is refused with a message.
What is lost is the audit path #36 asked for — the boundary readable without
reading the hooks.

The sweep half is the one worth the ticket. A guard whose header explains its
soundness by reference to a procedure nobody wrote is making a claim the
repository cannot honour. `no-work-on-stale-branch.sh` is careful about exactly
this elsewhere: it names `allow_squash_merge` as a setting its fallback depends
on and says plainly that the setting was not yet applied. The sweep gets no such
treatment — it is named as a thing that happens.

And #36 cannot be closed as delivered-as-amended while its own Amendment's only
re-scope is outstanding in both halves.

## Suggested fix

**The vocabulary.** Extend `CONTEXT.md`'s *Worktree branch* entry with the
lifetime, in the Amendment's own terms: the branch exists for exactly one pull
request; when that pull request merges into the active dev branch the branch is
finished, removed remotely by `delete_branch_on_merge` and locally by the sweep;
a new unit of work starts on a new branch forked from the active dev branch's
tip, and an existing worktree is not reused. Then the two pointers into the
glossary become true.

`CLAUDE.md:239` already says the same thing in one sentence and should stay as
it is — the operative paragraph, not the definition.

**The sweep needs a decision before it needs an edit**, and the decision is
Bertan's, because both answers are defensible:

1. *Write it.* `branch-hygiene` gains a sweep section — local deletion of
   merged worktree branches and removal of their worktrees, run off
   `report-stale-branches.sh`'s output, with the same `-d`-never-`-D` care the
   rotation already takes and the unclassified case left alone. Removing a
   worktree is a reserved act, so this belongs in Bertan's procedure and not the
   agent's half.
2. *Retract the claim.* If the local half is in fact ad hoc, the honest fix is
   for both hook headers to say so — the remote half is automatic, the local
   half is manual and unscheduled — in the register the file already uses for
   `allow_squash_merge`.

What is not defensible is the present state, where the headers assert a
procedure and the skill does not contain one.

**The fifth reserved act** is independent of that decision and can land with the
vocabulary: add removing a worktree or deleting a worktree branch to
`CONTEXT.md`'s enumeration, so the two citations in `report-stale-branches.sh`
land somewhere. If `branch-hygiene`'s "four acts" phrasing is kept, it becomes
five.

## Notes

`ready-for-agent` is deliberately withheld. The vocabulary half is fully
specified and an agent could land it today; the sweep half turns on the decision
above, which is not an agent's to take. Split the ticket or take the decision —
either unblocks it.

## Evidence

- #36 body, "Amendment — 2026-09-09", the *Stage 3's scope changes* paragraph
- #41 `lastEditedAt` `2026-09-09T11:37:32Z`; its two `userContentEdits` diff only
  in the `Blocked by` line
- PR #48 files: `check-hooks.sh`, `branch-hygiene/SKILL.md`, `CLAUDE.md`,
  `CONTEXT.md` — merged `2026-09-09T14:46:35Z`
- PR #57 files: `check-hooks.sh`, `no-work-on-stale-branch.sh`,
  `report-stale-branches.sh`, `settings.json`, `branch-hygiene/SKILL.md` —
  merged `2026-09-09T18:05:53Z`
- `CONTEXT.md`, *Worktree branch* entry — no lifetime
- `.claude/hooks/no-work-on-stale-branch.sh:2-4`; `report-stale-branches.sh:2,21`
- `report-stale-branches.sh:19-20,159` against `CONTEXT.md`'s four-item
  *reserved act* list, and `branch-hygiene/SKILL.md:19` reading it as closed
- `git log --oneline -S "lives as long as its pull request" -- CLAUDE.md` → one
  commit, `765a637` (PR #66, issue #63, 2026-09-11)
- `grep -rn "worktree remove" CLAUDE.md CONTEXT.md .claude/skills/` → no matches

bgunyel:
Closes #70.

#36's Amendment re-scoped stage 3 in one sentence and neither half arrived.
This lands both, plus the one-line fix that rides with them.

## The vocabulary

`CONTEXT.md`'s *Worktree branch* entry defined the branch and stopped. A
reader following `no-work-on-stale-branch.sh`'s pointer learned everything
about a worktree branch except when it ends — which is the fact the refusal
turns on. The entry now carries the lifetime: one pull request, then the remote
half `delete_branch_on_merge` takes and the local half the sweep does, and the
worktree that produced the branch is not reused. That last clause is the
operative one; a worktree outliving its branch is precisely how work gets
committed onto a branch whose pull request has already merged.

`CLAUDE.md:239` is untouched, as #70 asked — the operative paragraph, not the
definition.

## The sweep

Named in two hook headers and written nowhere. #70 left the choice between
writing it and retracting the claim to Bertan, who chose **write it, and name
its cadence**. That decision and its rejected alternative are recorded on
[#70](https://github.com/bgunyel/clause-and-effect/issues/70#issuecomment-5638004356),
so the audit path does not depend on a transcript.

`branch-hygiene` gains it as a procedure of its own:

- **unlock, then remove, then delete** — see the review finding below;
- `-d` never `-D`, with the reason the rotation already gives, and the same
  squash-or-rebase exception;
- every `unclassified` branch left alone, since sweeping one deletes work that
  was about to start;
- **cadence stated outright** — manual and unscheduled, in the register
  `no-work-on-stale-branch.sh` already uses for `allow_squash_merge`. That is
  the half that makes the two headers honest: #70's complaint was not that the
  sweep was undocumented but that it "is named as a thing that happens".

Both headers now point at it rather than restating it, and the guard carries
only the consequence for itself — the window it covers opens at the merge and
closes when someone gets round to sweeping, so it is bounded by nothing.

## The fifth reserved act

Removing a worktree or deleting a worktree branch joins `CONTEXT.md`'s
enumeration, so the two citations in `report-stale-branches.sh` — its header,
and the line it prints into every session's transcript — land somewhere.

The count came **out** of `branch-hygiene` rather than being corrected from
four to five: a corrected count is the same defect with a later expiry date.
`CONTEXT.md` holds the list once and the skill cites it. Steps 2 **and** 3 of
the sweep are reserved, step 3 included rather than argued about, since
`git worktree prune` removes worktree entries too.

## The `CLAUDE.md` *Domain docs* hunk

Called out because a pull request this careful about each hunk should not leave
one unexplained. The sentence said the glossary held "so far the one
distinction between a *check* and a *probe*" while `CONTEXT.md` had five
entries — the `four acts` staleness in a second place, and already stale before
this branch touched anything. It no longer enumerates at all: a longer
enumeration guarded by nothing would go the same way.

## What guards it

`check-hooks.sh` gains twenty-one checks: the boundary-section check with its
direction reversed. There `settings.json` is the fact and `CLAUDE.md` the
claim; here the hooks are the fact — they ship, they run, their headers send a
reader somewhere — and the documents are what must be there on arrival.

Measured against the final check set, on a copy rather than in the tree:
**eleven red** with the three documents reverted to `dev-05` and the two hook
headers held at this branch, **thirteen** with the headers reverted too. (The
"ten of seventeen" this branch quoted earlier was a true measurement of the
first commit's check set and went stale when the second reworked it.)

### The second commit is a review finding against the first

The sweep-citation check originally asked whether each header contained
`sweep`. Both contained it at `dev-05` already — once in the guard, three times
in the report — and that bare word *is* the dangling citation #70 exists about.
So the check was satisfied by the defect and would have stayed green through a
revert of every line the two headers gained. The literal is now the pointer,
which appears in neither header at `dev-05`.

### The third commit is a review finding against the second

**The sweep as written could not run on this repository.** `EnterWorktree`
locks every worktree it creates; `git worktree remove` refuses a locked one and
offers `remove -f -f`. Worse, step 3 failed *silently*: a locked worktree is
exempt from pruning by design, so `git worktree prune` skips it, exits 0, and
`git worktree list` goes on naming it — the closing invariant never reached,
with nothing saying why.

Step 2 now unlocks first and sends the reader to the lock reason rather than to
git's suggested force. The reason carries the session name, the pid, and that
process's start time — field 22 of `/proc/<pid>/stat`, verified against a live
lock — which is what separates a live session from a stale lock whose pid has
been reused.

**The suite now names its own limit**, because that limit is what let this
ship: every literal in the section asks whether the section *contains* a
command. None runs one, so none is evidence the procedure succeeds. A sweep
naming three commands that all refuse would pass every check there.

## Verification

- `bash .claude/hooks/check-hooks.sh` — 624 checks, all pass.
- `make test` — **575 passed, 5 xfailed, 0 failed**, after `uv sync
  --all-groups`. The earlier `test_installed_packages_match_uv_lock` failure
  was this worktree's `.venv` short of `alembic` and `mako`; it is gone.
- `make verify` and `make upgrade-safe` **both fail, for reasons that predate
  this branch and that it cannot fix** — this diff changes no dependency
  (`git diff origin/dev-05...HEAD -- uv.lock pyproject.toml` is empty):
  - `make audit` fails on the **committed** lock: `accelerate==1.14.0`,
    PYSEC-2026-3804 / GHSA-4j2p-28q2-5m79, CVSS 7.1, and osv-scanner reports
    *0 vulnerabilities can be fixed* — no fixed version exists.
  - `make upgrade-safe` reverts the candidate lock and blocks on four GuardDog
    findings in upgrade candidates: `docling-slim==2.126.0`,
    `google-genai==2.23.0`, `huggingface-hub==1.31.0`, `transformers==5.17.0`.
    These need the `waiver-review` skill; accepting a finding writes to a
    machine-wide `accepted.json`, which is not something to do to unblock a
    documents-only pull request.

  `CLAUDE.md:32` makes `upgrade-safe` the floor before a pull request closes and
  states no exception for documents-only diffs. Flagging rather than waiving.

https://claude.ai/code/session_01Uke8GuhacrPbLvjPXWjmZr

### DIFF
--- a/.claude/hooks/check-hooks.sh
+++ b/.claude/hooks/check-hooks.sh
@@ -2430,6 +2430,24 @@ unarmed 'and it is that section rather than the rotation beside it' \
 # before #70 -- and it takes the rotation's care over the same flag.
 written 'the sweep removes the worktree' "$SWEEP_SECTION" 'git worktree remove'
 written 'and deletes the local branch' "$SWEEP_SECTION" 'git branch -d'
+# And unlocks it first, without which the other two cannot run here at all.
+# Review of the first two commits found the procedure unable to execute on this
+# repository: EnterWorktree locks every worktree it creates, `git worktree
+# remove` refuses a locked one and names `remove -f -f` as the way out, and
+# `git worktree prune` is exempted from locked worktrees by design -- so it
+# skips one, exits 0, and the closing invariant is never reached with nothing
+# saying why. Reproduced: all four worktrees present at the time carried
+# `locked claude session <name> (pid N start T)`.
+#
+# THE LIMIT, which is the one that let that ship. Every literal in this section
+# asks whether the section CONTAINS a command. None of them runs one, so none is
+# evidence that the procedure succeeds -- a sweep naming three commands that all
+# refuse would pass every check here. What guards the difference is a person
+# running it; these hold the text against the citations, and nothing more.
+written 'and unlocks it first, which is what makes the other two possible' \
+  "$SWEEP_SECTION" 'git worktree unlock'
+written 'and warns that prune will not rescue a worktree still locked' \
+  "$SWEEP_SECTION" 'exempt from pruning by design'
 written 'with the care the rotation takes over the same flag' \
   "$SWEEP_SECTION" 'never `-D`'
 # The report classifies three ways and only one of the three is the sweep's. A
--- a/.claude/hooks/check-hooks.sh
+++ b/.claude/hooks/check-hooks.sh
@@ -2311,6 +2311,108 @@ tok 'one hook runs on an Edit, which is the number the section claims' \
                 | $m == "*" or $m == "" or ("Edit" | test($m)))
               | .hooks[]?] | length' "$SETTINGS" 2>/dev/null)"
 
+echo "=== the documents answer the citations the hooks make into them ==="
+# A fourth kind of check, and the section above with its direction reversed:
+# there settings.json is the fact and CLAUDE.md the claim; here the hooks are
+# the fact -- they ship, they run, and their headers send a reader somewhere --
+# and the documents are what has to be there when the reader arrives. Issue #70
+# found three such citations landing nowhere. Each had shipped; each is read at
+# the moment a reader has just been refused something; and none of them was
+# wrong in a way this suite could see, because nothing held a hook's pointer to
+# the thing it points at.
+#
+# These are evidence about the three citations named below and nothing else. A
+# fourth pointer added to a hook tomorrow is uncounted here, and so is any of
+# these three reworded, because every literal is the sentence as written.
+#
+# CONTEXT.md's entries are extracted rather than grepped whole, for the reason
+# the boundary paragraph is narrowed above: a glossary-wide grep is satisfied by
+# the word turning up in a neighbouring entry, which is the failure being fixed
+# rather than a check on it -- #70's finding was that the lifetime rule was
+# written down twice, in neither place a reader looking for vocabulary would go.
+# An entry runs from its bolded name to its `_Avoid_:` line, and the extraction
+# is checked from both ends before anything is asserted against it.
+CONTEXT_MD="$HOOKS/../../CONTEXT.md"
+SKILL_MD="$HOOKS/../skills/branch-hygiene/SKILL.md"
+entry() {  # entry <file> <bolded name> -- one glossary entry, name to _Avoid_
+  awk -v name="**$2**:" '$0 == name {f=1} f {print} f && /^_Avoid_:/ {exit}' "$1" 2>/dev/null
+}
+
+WORKTREE_ENTRY="$FIXTURES/context-worktree-branch.md"
+entry "$CONTEXT_MD" 'Worktree branch' > "$WORKTREE_ENTRY"
+written 'the extracted entry is the worktree branch entry' \
+  "$WORKTREE_ENTRY" '**Worktree branch**:'
+unarmed 'and it is that entry rather than the whole glossary' \
+  "$WORKTREE_ENTRY" '**Reserved act**:'
+
+# no-work-on-stale-branch.sh sends a reader here for what a worktree branch is,
+# and what that hook refuses is a commit on one whose pull request has merged.
+# Before #70 the entry defined the branch and stopped, so a reader who followed
+# the pointer learned everything about it except the fact the refusal turns on.
+written 'the entry says the branch lives for one pull request' \
+  "$WORKTREE_ENTRY" 'exists for exactly one pull request'
+written 'and that the worktree it was made in is not reused after it' \
+  "$WORKTREE_ENTRY" 'is not reused'
+
+RESERVED_ENTRY="$FIXTURES/context-reserved-act.md"
+entry "$CONTEXT_MD" 'Reserved act' > "$RESERVED_ENTRY"
+written 'the extracted entry is the reserved act entry' \
+  "$RESERVED_ENTRY" '**Reserved act**:'
+unarmed 'and it is that entry rather than the whole glossary' \
+  "$RESERVED_ENTRY" '**Worktree branch**:'
+
+# report-stale-branches.sh calls removing a worktree "a reserved act in
+# CONTEXT.md" in its header, and prints the same claim into every session's
+# transcript. The enumeration named four acts and that was not one of them.
+written 'the enumeration names the act the report cites' \
+  "$RESERVED_ENTRY" 'removing a worktree or deleting a worktree branch'
+
+# The skill reads that enumeration as closed and counts it. An act added to
+# CONTEXT.md and not to the sentence that counts them leaves the skill asserting
+# a number the glossary has outgrown -- the drift the section above is checked
+# for, one document along. Both directions, because only the second catches a
+# count corrected by deleting the claim instead of fixing it.
+written 'the skill counts the acts as the glossary now enumerates them' \
+  "$SKILL_MD" 'five acts'
+unarmed 'and does not still call the enumeration four' \
+  "$SKILL_MD" 'four acts'
+
+# The third citation, and the one that had gone unwritten rather than merely
+# undocumented: both hooks name a local sweep as what removes a merged worktree
+# branch, and no such procedure existed. The hooks are read for the citation
+# and the skill asserted to answer it, rather than the sweep's existence being
+# asserted on its own -- a procedure nothing cites is a procedure that can go.
+for hook in no-work-on-stale-branch.sh report-stale-branches.sh; do
+  tok "$hook cites a local sweep" 'cited' \
+      "$([ "$(prose_count "$HOOKS/$hook" 'sweep')" -gt 0 ] && echo cited || echo absent)"
+done
+
+# Extracted for the reason the glossary entries are: `git branch -d` is in this
+# file already, in the rotation, so a file-wide grep for it would pass with no
+# sweep written at all. Stopping at the next `## ` and not at the next heading
+# of any depth, because the sweep is numbered into steps the way the rotation
+# is -- a stop on `### ` would end the section at its own first step and assert
+# the rest of it against nothing.
+SWEEP_SECTION="$FIXTURES/branch-hygiene-sweep.md"
+awk '/^## The sweep/ {f=1; print; next} f && /^## / {exit} f {print}' \
+    "$SKILL_MD" > "$SWEEP_SECTION"
+written 'the extracted section is the sweep' "$SWEEP_SECTION" 'The sweep'
+unarmed 'and it is that section rather than the rotation beside it' \
+  "$SWEEP_SECTION" 'Create the new branch'
+
+# What a sweep has to say to be the thing those two headers name: it deletes the
+# local branch, it removes the worktree standing on it -- the act CONTEXT.md now
+# reserves, and one that no command in this repository's documents performed
+# before #70 -- and it takes the rotation's care over the same flag.
+written 'the sweep removes the worktree' "$SWEEP_SECTION" 'git worktree remove'
+written 'and deletes the local branch' "$SWEEP_SECTION" 'git branch -d'
+written 'with the care the rotation takes over the same flag' \
+  "$SWEEP_SECTION" 'never `-D`'
+# The report classifies three ways and only one of the three is the sweep's. A
+# sweep that acted on `unclassified` would delete a branch freshly cut for work
+# not yet started, which is the case that classification exists to protect.
+written 'and leaves the unclassified alone' "$SWEEP_SECTION" 'unclassified'
+
 echo
 if [ $FAILED -eq 0 ]; then echo "ALL CHECKS PASSED"; else echo "SOME CHECKS FAILED"; fi
 exit $FAILED
--- a/.claude/hooks/check-hooks.sh
+++ b/.claude/hooks/check-hooks.sh
@@ -2331,7 +2331,13 @@ echo "=== the documents answer the citations the hooks make into them ==="
 # rather than a check on it -- #70's finding was that the lifetime rule was
 # written down twice, in neither place a reader looking for vocabulary would go.
 # An entry runs from its bolded name to its `_Avoid_:` line, and the extraction
-# is checked from both ends before anything is asserted against it.
+# is checked from both ends before anything is asserted against it -- with one
+# limit named, because the two extractions below are not equally evidenced.
+# *Reserved act* has an entry after it, so its `unarmed` is real evidence that
+# the `_Avoid_:` stop fires. *Worktree branch* is the last entry in the file:
+# nothing follows it for an over-run to swallow, so its `unarmed` tests only
+# that the extraction did not begin too early, and the `_Avoid_:` stop is
+# evidenced there by the other extraction rather than by its own.
 CONTEXT_MD="$HOOKS/../../CONTEXT.md"
 SKILL_MD="$HOOKS/../skills/branch-hygiene/SKILL.md"
 entry() {  # entry <file> <bolded name> -- one glossary entry, name to _Avoid_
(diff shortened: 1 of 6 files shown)


bgunyel:
## Review against #70

Every claim below was re-run rather than read. Measured on `29896eb`.

### The acceptance criteria are met

| #70 asked for | |
|---|---|
| `CONTEXT.md`'s *Worktree branch* entry gains the lifetime — exactly one pull request; finished when it merges into the active dev branch; remote half `delete_branch_on_merge`, local half the sweep; the next unit cut from the dev tip; **the worktree not reused** | ✅ all six |
| `CLAUDE.md:239` stays as it is | ✅ `git diff origin/dev-05...HEAD -- CLAUDE.md` contains no hit for `lives as long as its pull request` |
| The sweep written: run off the report, `-d` never `-D`, `unclassified` left alone, Bertan's rather than the agent's | ✅ all four, plus the cadence |
| The fifth reserved act added to the enumeration | ✅ — and taking the count **out** of `branch-hygiene` rather than correcting it to five is the better reading of the ticket than the ticket's own suggestion |

### What was verified

- `bash .claude/hooks/check-hooks.sh` — 622 checks, all pass.
- `make test` — 574 passed, 5 xfailed, one failure: `test_installed_packages_match_uv_lock`, this worktree's `.venv` short of `alembic` and `mako`. Reproduced, and unrelated — the diff contains no Python.
- `delete_branch_on_merge` is in fact `true` on this repository (`gh api repos/bgunyel/clause-and-effect`), so the new glossary sentence asserts something true rather than something intended.
- No `four acts`, no `five acts`, no `Bertan's sweep` survives anywhere in the tree.
- **The nineteen checks are armed.** Reverting only `CONTEXT.md`, `CLAUDE.md` and `SKILL.md` to `dev-05` and keeping the two hook headers turns **eleven** of them red; reverting all six turns thirteen red. The second commit's finding is sound: the pointer literal appears in neither header at `dev-05`, so the `sweep` version really would have stayed green through a revert.
- Step 2's ordering claim is right, reproduced verbatim: `error: cannot delete branch 'f' used by worktree at '<path>'`.

### One finding: the sweep cannot run on this repository's worktrees

Every worktree a Claude Code session creates is `git worktree lock`ed. All four present right now are locked, by live sessions.

Step 2 fails:

```
$ git worktree remove ../wt
fatal: cannot remove a locked working tree, lock reason: claude session test
use 'remove -f -f' to override or unlock first
$ echo $?
128
```

Step 3 fails **silently**, which is worse:

```
$ rm -rf ../wt && git worktree prune -v ; echo "exit=$?"
exit=0
$ git worktree list
.../repo  57cdbd9 [master]
.../wt    57cdbd9 [f] locked
```

`prune` skips a locked worktree and reports success, so the procedure's closing invariant — "`git worktree list` names no worktree without a branch" — is not reached and nothing says so. `git branch -d` then still refuses.

The procedure needs a `git worktree unlock <path>` step, with the same *look before you force* treatment the `--force` paragraph already gives, rather than a reader arriving at `remove -f -f` from git's own error text.

This is the defect class #70 exists about — a documented procedure the repository cannot perform — and it is the one thing the new checks structurally cannot catch, since they assert that the section contains the commands, not that the commands succeed.

In fairness: a session that ends cleanly appears to remove its own worktree (one did so while this review was running). The lock therefore bites on worktrees kept past session end, which is exactly the population the sweep exists for.

### Minor

1. **Step 3 changes things too.** `SKILL.md` reserves "the whole of the rotation below, and step 2 of the sweep", but step 3 runs `git worktree prune` — removing worktree administrative entries, arguably the act `CONTEXT.md` now reserves — and `git fetch --prune`. Either step 3 is reserved as well, or the sentence should say why pruning metadata is not the reserved act.
2. **"Ten were red before the document edits" is eleven.** Counted with the two hook headers held at this branch's version; thirteen with them reverted too.
3. **#70's decision is not recorded.** This rests on "#70 left the choice ... to Bertan, who chose **write it, and name its cadence**". #70 has no comments, `createdAt == updatedAt`, and still carries only `documentation` — `ready-for-agent` withheld exactly as its Notes said. If the decision was taken in a session, the audit path #36 asked for does not have it. One comment on #70 closes that.
4. **`make upgrade-safe` has not been run**, and `CLAUDE.md:32` states the rule without an exception for documents-only diffs. Declared honestly in the body; `uv sync --all-groups` is needed first, which is what the one test failure is.
5. **The `CLAUDE.md` *Domain docs* hunk is not described in the body.** It is legitimate — the old sentence called the glossary "so far the one distinction between a *check* and a *probe*" and there are now five entries, which is the `four acts` staleness in a second place — but a pull request this careful to explain each hunk should say so.

### Verdict

Approve once the unlock step is in. Nothing above touches the substance of what #70 asked for, which is here in full.

The pull request's worktree was not touched; a live session holds it.

https://claude.ai/code/session_012Wnj3VfjC4HBvzR87c3H9M

bgunyel:
## Addressed in `c68e318`

Every number below was re-measured here rather than taken from the review.

### The finding: the sweep cannot run on this repository's worktrees

Confirmed, and worse than a missing command. Reproduced: all three worktrees
present carry `locked claude session <name> (pid N start T)`, and git's own
documentation says the quiet half outright — locking exists so that "you can
prevent its administrative files from being pruned". So `prune`'s skip is by
design, not a bug, which is exactly why it exits 0 and says nothing.

Step 2 now unlocks first. It sends the reader to the lock reason rather than to
the `remove -f -f` git offers in its own error text, and the reason turns out to
carry enough to decide: session name, pid, and that process's start time. The
start time is field 22 of `/proc/<pid>/stat` — checked against a live lock here
rather than assumed — which is what separates a live session from a stale lock
whose pid has since been handed to something else. The procedure confirms the
process is gone rather than inferring it from the branch being merged.

Step 3 now says what it cannot do, and the closing invariant is explicitly
checked against `git worktree list` rather than inferred from the exit codes
above it.

Two checks added, and **armed** rather than assumed: measured red against
`29896eb`, which had the sweep and not the unlock.

On the point that the checks structurally cannot catch this — agreed, and it is
now written into the suite as its own limit, because it is what let the defect
ship:

> Every literal in this section asks whether the section CONTAINS a command.
> None of them runs one, so none is evidence that the procedure succeeds — a
> sweep naming three commands that all refuse would pass every check here.

### Minor 1 — step 3 changes things too

Taken. Reserved now reads "steps 2 and 3 of the sweep", with the reason for not
arguing the distinction: `git worktree prune` removes worktree entries, and
whether pruning an administrative file is *really* removing a worktree is
something a reader would have to get right unprompted, for no gain over
reserving both.

### Minor 2 — the count

You are right, and the cause is worth naming: "ten of seventeen" was a true
measurement of the **first** commit's check set, and went stale when the second
commit reworked those checks — the branch's own staleness defect, in its own
commit message. Re-measured against the twenty-one checks now present, on a
copy rather than in the tree: **eleven** with the three documents reverted and
the headers held here, **thirteen** with the headers reverted too. Both figures
match yours exactly. Corrected in the pull request body.

### Minor 3 — the decision is not recorded

Fixed: [#70 comment](https://github.com/bgunyel/clause-and-effect/issues/70#issuecomment-5638004356),
recording the choice, the rejected alternative, and why the rejected one would
have been worse than it looked — retracting the claim would never have exposed
the lock defect, because nobody would have tried to run the procedure.

It also flags, rather than hides, that `CLAUDE.md`'s documentation table has a
reasonable claim on this: a decision with a rejected alternative belongs in
`docs/adr/`. No ADR written — #70 did not ask for one and it is a heavier act
than this ticket scoped. The comment is its raw material if you want it.

### Minor 4 — `make upgrade-safe`

Run now, and it fails — along with `make verify` — for reasons that predate this
branch and that it cannot fix. This diff changes no dependency
(`git diff origin/dev-05...HEAD -- uv.lock pyproject.toml` is empty):

- **`make audit`, on the committed lock:** `accelerate==1.14.0`,
  PYSEC-2026-3804 / GHSA-4j2p-28q2-5m79, CVSS 7.1 — and osv-scanner reports
  *0 vulnerabilities can be fixed*. There is no fixed version to move to.
- **`make upgrade-safe`:** reverted the candidate lock cleanly and blocked on
  four GuardDog findings in upgrade candidates — `docling-slim==2.126.0`,
  `google-genai==2.23.0`, `huggingface-hub==1.31.0`, `transformers==5.17.0`.

Not waived. Those need `waiver-review`, and accepting a finding writes to a
machine-wide `accepted.json` shared beyond this repository — waiving a security
finding to unblock a documents-only pull request is the wrong direction for
that incentive to run. `CLAUDE.md:32` states the rule with no exception for
documents-only diffs, so this is yours to decide rather than mine to route
around.

One consequence worth stating plainly: as things stand **no** pull request can
satisfy that gate, so it is not a bar this branch can clear by trying harder.

`uv sync --all-groups` was run, and `make test` is now **575 passed, 5 xfailed,
0 failed**.

### Minor 5 — the `CLAUDE.md` hunk

Described in the body now, under its own heading.

### Note

`dev-05` moved while this was in review — `CLAUDE.md`'s *Deliberately left
open* section went from four consequences to five, in the main checkout. It is
a different section from this branch's *Domain docs* hunk, so it should merge
clean, but flagging it since it is your change and not mine to touch.

https://claude.ai/code/session_01Uke8GuhacrPbLvjPXWjmZr

bgunyel:
`dev-05` moved while the reply above was being written — #75 and #76 landed —
and this went `CONFLICTING`. Merged in `a57bac9`; now `MERGEABLE` / `CLEAN`.

One conflict, in `check-hooks.sh`, additive on both sides: each branch appended
a block at the end. Both kept, and the order is not arbitrary. #76's left-open
count check opens no `echo` heading of its own and reads the `$SECTION` fixture
built by *CLAUDE.md names every hook that carries the boundary*, so it belongs
inside that section and goes first; this branch's *the documents answer the
citations* section opens its own heading and follows. Reversed, #76's check
would have been filed under a heading with nothing to do with it — still green,
and misleading to the next reader.

`CLAUDE.md` merged clean: #76 rewrote *Deliberately left open*, this branch
rewrote *Domain docs*, and they do not touch.

Re-verified after the merge rather than assumed:

- `check-hooks.sh` — 624 checks, all pass, **including #76's new count check**
  against the merged `CLAUDE.md`.
- `make test` — 586 passed, 5 xfailed, 0 failed.

The two gate failures in the reply above are unchanged and unrelated to the
merge: `accelerate==1.14.0` with no fixed version, and four GuardDog blockers on
upgrade candidates. Still yours to decide, still not waived.

PR #81, opened 2026-09-11
Match `gh` as a word, so a commit message saying *high* is not a PR decision

### LINKED ISSUE #72: GH_SURFACE_ANYWHERE matches `gh` as a substring: a wrapped commit whose message says *high* is refused as a PR decision
## Parent

#36

## Context

Found by review on `dev-05`, carried over unfixed in the spec-verification
handoff of 2026-09-09 as the first of three known-open items. Every verdict
below was re-measured against the hooks as they stand on `main` at `7cb4891`.
`0` = allow, `2` = block.

This is the boundary-hook half of the defect class #69 reports in the
convention hooks: a command name matched as a substring rather than as a word.
#69 is explicitly scoped away from the boundary hooks, so this is not covered
there. It is a different mechanism from #68 — the wrapper rules never reach
`cs_split`, they read `WRAPTEXT` — so it is not covered there either.

Over-refusal only. Nothing is permitted that was not permitted before.

## What is wrong

`no-pr-decisions.sh:352`:

```sh
GH_SURFACE_ANYWHERE='gh[[:space:]]+(.*[^-A-Za-z0-9_])?(pr|release|api)([^-A-Za-z0-9_]|$)'
```

`gh` has no left boundary. Any word *ending* in `gh` satisfies it, so *high*,
*enough*, *through*, *though*, *rough*, *sigh* and *dough* are each a `gh` as
far as this rule is concerned. Once a wrapper is on the line, one of those
words followed by whitespace and, anywhere later, a delimited `pr`, `release`
or `api` is refused.

Every other token in the same expression is bounded, and each boundary is
argued for at length a few lines above it:

- the group's right edge, `([^-A-Za-z0-9_]|$)`, in this same pattern;
- `eval([^-A-Za-z0-9_]|$)` in the wrapper detector on the line that uses it;
- `release delete` against `release delete-asset`, justified as "a path word is
  matched whole";
- `rest_bases` anchoring on the field flag, justified as what "keeps `rebase`
  and `database` the words they are".

`gh` is the one token in the file that is not matched whole, and it is the only
boundary decision here that no comment defends. A deliberate asymmetry in a
file commented this heavily would have a paragraph.

## It contradicts what the suite says the rule does

`check-hooks.sh:831-833`, written directly above the wrapper ALLOW fixtures:

> The rule reaches gh's three deciding surfaces and stops there. A wrapped
> command that is none of them is answered by whatever else covers it, and by
> this file not at all.

`bash -c "git commit -m 'refactor high level api client'"` is a wrapped command
that is none of gh's three deciding surfaces, and this file answers it. That
sentence is the spec; the pattern does not implement it.

Nor is this the trade the file takes knowingly. All three parts of it
presuppose `gh` on the line — "Every wrapped `gh pr`…", "a wrapped `gh` command
that is none of the three…", "an entirely unwrapped one that merely shares a
line with a wrapper". A command with no `gh` in it at all falls outside all
three. CLAUDE.md's **Deliberately left open** names four consequences and this
is not among them. The rule currently refuses more than any document in this
repository says it refuses.

The refusal text is also false rather than merely conservative — *"A shell
wrapper does not change what the command decides, and its payload cannot be
read"* — said to a `git commit` that decides nothing. Over-refusal with a true
explanation costs one edit; over-refusal with a false one costs the edit and a
wrong belief about the command.

## Evidence

| now | fixed | command |
| --- | --- | --- |
| **2** | 0 | `bash -c "git commit -m 'refactor high level api client'"` |
| **2** | 0 | `bash -c "grep high pr.txt"` |
| **2** | 0 | `bash -c "cat sigh api.md"` |
| **2** | 0 | `bash -c "ls dough api"` |
| **2** | 0 | `bash -c "echo through pr"` |
| **2** | 0 | `bash -c "cat enough release.md"` |

The first row is the one that matters. It contains no `gh`, calls no GitHub
surface, and is the shape of an ordinary commit from inside a wrapper. The
other five are the handoff's examples, and they are contrived enough that the
finding reads as a curiosity when they are the whole of it.

Controls, unchanged in both directions:

| now | fixed | command |
| --- | --- | --- |
| 0 | 0 | `bash -c "make test"` |
| 0 | 0 | `bash -c "gh issue list"` |
| 2 | 2 | `bash -c "gh pr merge 5"` |
| 2 | 2 | `bash -c "gh pr view 5"` |
| 2 | 2 | `bash -c "./gh pr merge 5"` |
| 2 | 2 | `bash -c "/usr/bin/gh pr merge 5"` |
| 2 | 2 | `bash -c "gh release create v1"` |
| 2 | 2 | `eval "gh pr merge 5"` |
| 2 | 2 | `bash -c "make test" && gh pr view 5` |

## Solution

Give `gh` the left boundary every other token here already has:

```sh
GH_SURFACE_ANYWHERE='(^|[^-A-Za-z0-9_])gh[[:space:]]+(.*[^-A-Za-z0-9_])?(pr|release|api)([^-A-Za-z0-9_]|$)'
```

The pattern is only ever used under `grep -qE` as a boolean, so consuming the
boundary character costs nothing.

### The narrowing this accepts, which the handoff did not record

The character class excludes `-` and `_`, consistent with the rest of the file,
so a command whose *name* ends in `gh` behind one of those stops matching:

| now | fixed | command |
| --- | --- | --- |
| 2 | **0** | `bash -c "my-gh pr merge 5"` |
| 2 | **0** | `bash -c "my_gh pr merge 5"` |

There is a second, subtler one. A literal `\n` escape puts a word character
immediately before `gh`, so a line that spells `printf '…\ngh pr review
--approve …'` stops matching. That is the single verdict the fix changes across
the whole measured corpus below, and it is a loosening rather than the false
refusal it was counted as.

Both are evasion shapes, not mistakes, and `./gh` and `/usr/bin/gh` still
refuse. Accepted under "these stop mistakes, not adversaries" — but stated here
rather than left for a later review to find.

### Checks are owed either way

The fix reddens nothing. That is not an argument against it — the justification
is the false refusal, which exists without the suite. What it establishes is
that **the suite has a hole in the permitting direction**, while CLAUDE.md
claims `check-hooks.sh` "checks the boundary in both directions".

The only ALLOW fixture guarding this path is `check-hooks.sh:836`,
`bash -c "make test"` — a command with no `gh`-ending word and no trigger
token, which cannot trip the pattern under any regex. Its two neighbours are
`gh issue close` and `gh issue list`, which exercise the subcommand dimension,
not this one. One better fixture would have caught this a review ago.

So:

- **If the fix lands**, add `ALLOW bash -c "git commit -m 'high level api client'"`
  and `BLOCK bash -c "./gh pr merge 5"`, pinning the behaviour in both
  directions, plus a row for the `-`/`_` narrowing.
- **If it does not**, the trade comment above `GH_SURFACE_ANYWHERE` and
  CLAUDE.md's **Deliberately left open** list need a fifth entry saying that a
  wrapped command containing no `gh` at all is refused when its prose happens
  to carry one of these words. Doing nothing and documenting nothing is the one
  option the repository's own stopping rule does not allow — a rule that
  refuses more than its comment says is a stale docstring, which this repo
  counts as a defect.

## Severity, measured

Low. Real, but not urgent, and it blocks nothing.

Against 7,621 real `Bash` commands taken from the local session transcripts of
this repository and every other project on this machine: 66 were
wrapper-detected, and the fix changes the verdict on exactly **1** — the
`printf '…\ngh pr review --approve…'` line described above, which is a
narrowing rather than a false refusal repaired.

The full suite was run twice, once unmodified and once against a throwaway
clone carrying only this one-line change, and every verdict line diffed:
**identical, 594 ok / 1 FAIL both times.**

Two numbers from the handoff to restate while this is open:

- the suite is **595 checks, not 550**;
- it is **not green on `main` today**. `the extracted section is the boundary
  section` fails, unrelated to this — it is the CLAUDE.md section-extraction
  check, and is the same class of claim #63 closed. Worth its own issue if it
  is not already someone's.

"Passes all 550 checks" should therefore be read as "changes no verdict in the
595-check suite, which has one pre-existing unrelated failure".

bgunyel:
Fixes the over-refusal reported in #72: `GH_SURFACE_ANYWHERE` matched `gh` as a
substring, so any word *ending* in gh was a gh.

## The rule

```diff
-GH_SURFACE_ANYWHERE='gh[[:space:]]+(.*[^-A-Za-z0-9_])?(pr|release|api)([^-A-Za-z0-9_]|$)'
+GH_SURFACE_ANYWHERE='(^|[^-A-Za-z0-9_])gh[[:space:]]+(.*[^-A-Za-z0-9_])?(pr|release|api)([^-A-Za-z0-9_]|$)'
```

The boundary is the one every other token in the file already has, each argued
for a few lines away — the group's own right edge, `eval` in the wrapper
detector, `release delete` against `release delete-asset`, `rest_bases`
anchoring on its field flag. `gh` was the only boundary decision here that no
comment defended. The pattern has one definition and one use, under `grep -qE`
as a boolean, so consuming the boundary character costs nothing.

Over-refusal only. Nothing is permitted that was not permitted before, except
the three evasion shapes recorded below.

## Why it mattered

`bash -c "git commit -m 'refactor high level api client'"` names no gh, calls no
GitHub surface, and decides nothing — and was refused as a PR decision, with a
reason that was false rather than merely conservative: *"A shell wrapper does
not change what the command decides, and its payload cannot be read."*

It was outside every trade this repository has written down. All three parts of
the trade comment presuppose a `gh` on the line, and CLAUDE.md's **Deliberately
left open** list names no such consequence. `check-hooks.sh` states the spec
directly — *"The rule reaches gh's three deciding surfaces and stops there"* —
and the pattern did not implement it.

## Measured

| | before | after |
|---|---|---|
| the 6 evidence rows in #72 | BLOCK | **ALLOW** |
| the 9 control rows in #72 | unchanged | unchanged |
| `./gh`, `/usr/bin/gh` | BLOCK | BLOCK |

## The trade this accepts, recorded rather than left to be found

Three shapes stop matching, with **two different causes** — kept apart in the
comment because only one is a choice this file made:

- The boundary **existing** narrows a literal `\n` escape written immediately
  before gh (the preceding character is then `n`). This is the single verdict
  the fix changes across the 7,621-command corpus #72 sampled, and it follows
  from having a boundary at all rather than from which one — measured against
  both candidate classes, which agree.
- The class **excluding** `-` and `_`, as the rest of the file does, separately
  narrows `my-gh pr merge 5` and `my_gh pr merge 5`.

All are evasion shapes, not mistakes; `./gh` and `/usr/bin/gh` still refuse.
Accepted under *"these stop mistakes, not adversaries"*.

## Checks — 11 rows, because the suite had a hole in the permitting direction

CLAUDE.md claims `check-hooks.sh` "checks the boundary in both directions". The
only ALLOW guarding this path was `bash -c "make test"`, which carries neither a
gh-ending word nor a trigger token and so could not have tripped the pattern
under any regex. The new rows pin all three directions: the 6 false refusals,
the 2 path spellings that must still refuse, and the 3 shapes the boundary gives
up.

**Mutation-tested three ways** — the repo's rule is that no fix lands without a
check that fails without it:

| mutation | rows failed | of which new |
|---|---|---|
| restore the defect (no left boundary) | 9 | **9** |
| left *anchor* `(^\|[[:space:]])` instead of a class | 33 | 2 (`./gh`, `/usr/bin/gh`) |
| class widened to admit `-` and `_` | 2 | **2** — invisible to the suite as it stood |

The third mutation is the one that matters: without the new rows nothing in the
suite would have noticed it.

## Review

Both axes of `/code-review` ran. Spec: no findings. Standards: three findings,
all about what the new comments claimed rather than about the rule — a cause
mis-attributed, a "pinned in both directions" that covered two of three shapes,
and a comment pointing at its subject by line count. All three fixed in 698de8f;
the third was fixed by adding the missing row rather than by narrowing the
claim.

## State

- `bash .claude/hooks/check-hooks.sh` — **629 checks, all pass.** The failure #72
  records as pre-existing on `main` is already gone on `dev-05` (8b14525).
- `make test` — 586 passed, 5 recorded xfails.
- **No CLAUDE.md change is owed.** The "if the fix does not land" branch of #72
  asked for a new **Deliberately left open** entry; the fix landed. The fix also
  makes CLAUDE.md's existing "the `pr|release|api` group after a `gh`" true,
  where before it was not.
- No dev-log entry, matching the directly comparable hook PR (#76 / 8b14525).

Refs #72

https://claude.ai/code/session_01DuX7jC6CXMC8SB2PHC1oWb

### DIFF
--- a/.claude/hooks/check-hooks.sh
+++ b/.claude/hooks/check-hooks.sh
@@ -931,6 +931,46 @@ check no-pr-decisions.sh ALLOW 'bash -c gh issue close'      'bash -c "gh issue 
 check no-pr-decisions.sh ALLOW 'bash -c gh issue list'       'bash -c "gh issue list"'
 check no-pr-decisions.sh ALLOW 'bash -c an ordinary command' 'bash -c "make test"'
 
+echo "=== REGRESSION: #72, gh is a word here and not a suffix ==="
+# The sentence three lines above is the spec, and the pattern did not implement
+# it. `gh` was the one token in this file matched unbounded on its left, so any
+# word ENDING in gh satisfied it -- high, enough, through, sigh, dough -- and
+# once a wrapper was on the line, one of those followed by a delimited pr,
+# release or api anywhere later was refused. The first row is the one that
+# matters: it names no gh, calls no GitHub surface, and is the shape of an
+# ordinary commit from inside a wrapper. It was refused with a reason that was
+# false rather than merely conservative -- "a shell wrapper does not change what
+# the command decides" said to a commit that decides nothing.
+#
+# The ALLOW directly above was the only fixture guarding this path, and it
+# carries neither a gh-ending word nor a trigger token, so no regex could have
+# tripped it; its two neighbours exercise the subcommand dimension, not this
+# one. That is the hole this section fills. Measured before the one-line fix:
+# every ALLOW below was BLOCK.
+check no-pr-decisions.sh ALLOW 'bash -c a commit message saying high' \
+  "bash -c \"git commit -m 'refactor high level api client'\""
+check no-pr-decisions.sh ALLOW 'bash -c grep high pr.txt'      'bash -c "grep high pr.txt"'
+check no-pr-decisions.sh ALLOW 'bash -c cat sigh api.md'       'bash -c "cat sigh api.md"'
+check no-pr-decisions.sh ALLOW 'bash -c ls dough api'          'bash -c "ls dough api"'
+check no-pr-decisions.sh ALLOW 'bash -c echo through pr'       'bash -c "echo through pr"'
+check no-pr-decisions.sh ALLOW 'bash -c cat enough release.md' 'bash -c "cat enough release.md"'
+# A left boundary, not a left anchor. A path ends in a character that is none of
+# gh's own, so an invoked gh still matches wherever it is spelled from. Without
+# this pair the boundary could be tightened to `(^|[[:space:]])` -- or the whole
+# group anchored -- and nothing in this suite would notice.
+check no-pr-decisions.sh BLOCK 'bash -c ./gh pr merge'         'bash -c "./gh pr merge 5"'
+check no-pr-decisions.sh BLOCK 'bash -c /usr/bin/gh pr merge'  'bash -c "/usr/bin/gh pr merge 5"'
+# The narrowing the boundary accepts, pinned rather than left for a later review
+# to find. The character class excludes - and _ as every other token in this
+# file does, so a command whose NAME ends in gh behind one of those stops
+# matching. Both are evasion shapes rather than mistakes, and the two BLOCKs
+# above show the spellings that reach gh itself still refuse; accepted under
+# "these stop mistakes, not adversaries". Measured before the fix: both BLOCK.
+# This pair is what would notice if the trade were quietly taken back, or
+# quietly widened past what the comment on the pattern claims.
+check no-pr-decisions.sh ALLOW 'bash -c my-gh pr merge'        'bash -c "my-gh pr merge 5"'
+check no-pr-decisions.sh ALLOW 'bash -c my_gh pr merge'        'bash -c "my_gh pr merge 5"'
+
 echo "=== REGRESSION: review of 02a14d8, close and release through gh api ==="
 # Closing a PR and publishing a release were refused in the gh spelling and open
 # through gh api, so the boundary was spelling-dependent exactly where the file
--- a/.claude/hooks/check-hooks.sh
+++ b/.claude/hooks/check-hooks.sh
@@ -932,8 +932,10 @@ check no-pr-decisions.sh ALLOW 'bash -c gh issue list'       'bash -c "gh issue 
 check no-pr-decisions.sh ALLOW 'bash -c an ordinary command' 'bash -c "make test"'
 
 echo "=== REGRESSION: #72, gh is a word here and not a suffix ==="
-# The sentence three lines above is the spec, and the pattern did not implement
-# it. `gh` was the one token in this file matched unbounded on its left, so any
+# "The rule reaches gh's three deciding surfaces and stops there" -- the comment
+# heading the block above -- is the spec, and the pattern did not implement it.
+# (Named rather than pointed at by line count, which any insertion would make
+# wrong.) `gh` was the one token in this file matched unbounded on its left, so any
 # word ENDING in gh satisfied it -- high, enough, through, sigh, dough -- and
 # once a wrapper was on the line, one of those followed by a delimited pr,
 # release or api anywhere later was refused. The first row is the one that
@@ -970,6 +972,15 @@ check no-pr-decisions.sh BLOCK 'bash -c /usr/bin/gh pr merge'  'bash -c "/usr/bi
 # quietly widened past what the comment on the pattern claims.
 check no-pr-decisions.sh ALLOW 'bash -c my-gh pr merge'        'bash -c "my-gh pr merge 5"'
 check no-pr-decisions.sh ALLOW 'bash -c my_gh pr merge'        'bash -c "my_gh pr merge 5"'
+# The third narrowed shape, and the one with the other cause: a literal \n
+# escape puts `n` in front of gh, which no class this file would write admits,
+# so this follows from having a left boundary at all rather than from which one.
+# It is the single verdict the fix changed across the 7,621-command corpus #72
+# sampled. Pinned because the comment on the pattern claims all three are, and a
+# check suite is evidence about the cases it names and about nothing else -- the
+# two rows above cannot speak for this one, having a different cause.
+check no-pr-decisions.sh ALLOW 'bash -c a \n escape before gh' \
+  "bash -c \"printf 'summary\\ngh pr review --approve 5' > /tmp/x\""
 
 echo "=== REGRESSION: review of 02a14d8, close and release through gh api ==="
 # Closing a PR and publishing a release were refused in the gh spelling and open
--- a/.claude/hooks/no-pr-decisions.sh
+++ b/.claude/hooks/no-pr-decisions.sh
@@ -363,14 +363,24 @@ VERDICT='(^|[[:space:]])(--approve|--request-changes|-[A-Za-z]*[ar][A-Za-z]*)([[
 # under `grep -qE` as a boolean, so consuming the boundary character costs
 # nothing.
 #
-# The class excludes `-` and `_` as the rest of the file does, which narrows
-# three shapes: `my-gh pr merge 5` and `my_gh pr merge 5` stop matching, and so
-# does a literal `\n` escape written immediately before gh. All three are evasion
-# shapes rather than mistakes -- `./gh` and `/usr/bin/gh` still refuse, a path
-# ending in a character that is none of gh's own -- and they are accepted under
-# the same "these stop mistakes, not adversaries" that decides the rest, stated
-# here rather than left for a later review to find. Pinned in both directions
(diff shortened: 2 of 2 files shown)


bgunyel (commented):
## Review: acceptance criteria in #72 are met

Verified by measurement rather than by reading the description — every number below was re-derived locally, and where I could rebuild the evidence independently I did.

### The fix

The one-liner is character-for-character what #72 proposed. Worth stating the structural property the PR body leaves implicit: adding a **required** left boundary can only remove matches, so "over-refusal only, nothing newly permitted" holds by construction and not merely by measurement. The three narrowed shapes are the whole of what changes, and all three are pinned.

### Acceptance criteria

| criterion | result |
|---|---|
| 6 evidence rows BLOCK → ALLOW | ✅ all 6 |
| 9 control rows unchanged | ✅ all 9, incl. `./gh`, `/usr/bin/gh`, `eval`, shared-line |
| checks owed: ALLOW commit-message, BLOCK `./gh`, a `-`/`_` row | ✅ delivered as 11 rows — a superset |
| a fifth **Deliberately left open** entry | ✅ correctly **not** owed — that was the "if the fix does not land" branch |

The claim that the fix makes CLAUDE.md's existing *"the `pr|release|api` group after a `gh`"* true, where it was previously false, checks out.

### Mutation testing reproduces exactly

| mutation | rows failed | of which new |
|---|---|---|
| restore the defect | 9 | 9 |
| left *anchor* `(^\|[[:space:]])` | 33 | 2 |
| class widened to admit `-` and `_` | **2** | **2** |

The third is the one that carries the argument, and it holds: widening the class fails **only** the two new rows and leaves the `\n` row's verdict untouched — which is precisely why adding the row was the right call over narrowing the claim. The suite genuinely had a hole in the permitting direction that no existing fixture could have caught.

### The corpus claim, rebuilt independently

Rather than take the 7,621 figure on trust I rebuilt the corpus from 398 local transcripts: **8,687 commands, 8,046 unique** — a larger sample than the issue's. Of those, **101 were wrapper-detected, and exactly 1 verdict changes.** It is precisely the documented shape: a `printf` test fixture carrying `\ngh pr review --approve nonsense`, base `rc=2` → PR `rc=0`. Not a real `gh` invocation. The characterization in the comment is exact.

### Additional probing

Fourteen spellings not named in the PR — `;gh`, `&& gh`, `(gh`, `| gh`, tab, single-quote, `$(gh`, a **real** embedded newline, `gh release delete`, `gh api -X PUT /pulls/N/merge`, `gh pr review --approve`, `gh pr close` — all still BLOCK. The real-newline case is worth recording: `grep -E` anchors `^` per line, so an actual newline before `gh` still matches where the literal `\n` escape no longer does. The two are not the same shape, and only the escape is given up.

### One caveat, benign

The branch is **17 commits behind `dev-05`**, so every number in the description was measured against a stale base — including against the older `lib/command-scan.sh` that #74 rewrote (+194 lines), which is the library building the `WRAPTEXT` this pattern runs against. That is the one thing that could have invalidated the measurements, so I tested it: **the merge is clean, and the merged tree is green** — 728 checks all pass, `make test` 590 passed / 5 xfailed, all 17 behaviour rows identical, and the defect mutation still fails exactly the 9 new rows. No rework needed; merge-commit being the only method enabled here, merging will produce the tree I tested.

Unrelated: the `test_installed_packages_match_uv_lock` failure seen in a fresh worktree is environmental — it reproduces at the **base** commit, and this PR touches no dependency file.

### Not approving

Posting this as a comment rather than an approval on purpose: deciding a pull request is Bertan's call, which is the rule this very hook exists to enforce. The findings are above; the decision isn't mine to record.

https://claude.ai/code/session_014z6f4iGorSmUrHRAL8i3NS

Issue #79, opened 2026-09-11
The wrapper rules do not know the prefix words cs_split already strips: sudo sh -c evades every hook


bgunyel:
## Parent

#36

## Context

Found by review of #68 (PR #74), and confirmed unchanged by it — these verdicts
are identical before and after that branch. #68 explicitly scoped wrappers out,
so this is filed rather than folded into it.

## What is wrong

`cs_split` already maintains the list of words that run another command, in its
stage-2 prefix strip:

```
env|command|xargs|nohup|nice|time|stdbuf|ionice|sudo|doas|setsid|chronic   (+ timeout|flock)
```

The four wrapper regexes do not consult it. They answer "is this a wrapper that
runs another command?" independently and more narrowly — `(ba|z|)sh -c`,
`sh <<`, or `eval` — anchored at `^` or after a separator. That is the same
question answered in two places with two different answers, which is the defect
class the header of `lib/command-scan.sh` opens by naming.

The asymmetry it produces, measured:

```
BLOCK   sudo git push --all origin            cs_split strips sudo, the push is at a command position
ALLOW   sudo sh -c 'git push --all origin'    cs_split strips sudo, leaves sh -c '...', and the
                                              wrapper rule never fires because sh is not at ^
```

`sudo` is recognised in one half of the library and invisible in the other. The
same holds for `timeout`, `xargs`, `nohup` and `env`. It is one layer out from
the header's own PR #35 defect #2 — *"wrapper re-admission sufficed for a
heredoc and did nothing for `sh -c`"* — which now handles `sh -c` and does
nothing for `sudo sh -c`.

All ALLOW, measured against both `no-git-push.sh` and `no-commit-to-main.sh`:

```
sudo sh -c 'git push --all origin'
timeout 5 bash -c 'git push --all origin'
xargs sh -c 'git push --all origin'
env FOO=1 sh -c 'git push --all origin'
nohup sh -c 'git push --all origin'
find . -exec sh -c 'git push --all origin' \;
```

Note `env` rather than a bare assignment: `FOO=1 sh -c '...'` is already BLOCK,
because the anchor admits environment assignments. It is the wrapper *word*
that is missing, not the assignment.

## What to build

Extend the anchor to admit the prefix words the library already knows,
**reusing that list rather than restating it**. A second copy of those twelve
words in four hook regexes would be the same defect one more time.

## Two constraints, both load-bearing

**Do not simply drop the anchor.** It must widen to a known prefix set, not
disappear.

[...]

Issue #63, opened 2026-09-11
CLAUDE.md names two of the four hooks that carry the boundary


bgunyel:
## Parent

#36

## Context

Found by a two-axis review of `dev-05` against `main` (standards axis).

## What is wrong

CLAUDE.md's boundary section closes by naming where the boundary is enforced:

> Enforced by `.claude/hooks/no-git-push.sh` and `no-pr-decisions.sh`, both
> built on `.claude/hooks/lib/command-scan.sh`; `bash .claude/hooks/check-hooks.sh`
> checks the boundary in both directions.

Two hooks. `.claude/settings.json` registers **seven** Bash `PreToolUse` hooks:

| hook | named in the section? |
|---|---|
| `no-git-push.sh` | yes |
| `no-pr-decisions.sh` | yes |
| `no-commit-to-main.sh` | **no** |
| `no-work-on-stale-branch.sh` | **no** |
| `append-only-docs.sh` | no — different subject |
| `alembic-via-uv-group.sh` | no — different subject |
| `pytest-via-uv-group.sh` | no — different subject |

The last three are about documents and commands, not about the agent boundary,
and their absence is correct. The first two omissions are not:

- **`no-commit-to-main.sh`** enforces the section's own sentence, `main` and
  `dev-NN` are Bertan's to push. #43 rebuilt it on `lib/command-scan.sh`, so the
  section's "both built on" clause now describes three files and names two.
- **`no-work-on-stale-branch.sh`** is a boundary hook that refuses work, added
  by #44, and the section does not mention it at all — nor the branch-lifecycle
  rule it enforces, nor `report-stale-branches.sh` which arms it at
  `SessionStart`.

## Why it matters

The section exists so that the boundary can be audited without reading the
implementation. #36 states the requirement in its own words:

> As a reviewer of this repository, I want to know what an unattended agent can
> do to it without reading the two hooks and their tokeniser, so that I can
> judge the boundary rather than the implementation.

A reader who audits the two named files has audited less than half of what runs,
and has no way to discover that from the section. The failure is in the
direction that matters here: the text under-reports what is enforced, so a
reader concludes the boundary is narrower than it is and may "restore" a rule
that is already carried elsewhere — or, reading the same text as complete, may
conclude a refusal came from nowhere.

CLAUDE.md's own standard applies:

> a stale docstring here is a defect

The section was accurate when written. #43 and #44 landed after it and neither
revised it, which is the ordinary way this kind of claim goes stale.

[...]

PR #87, opened 2026-09-11
Convention hooks: close two silent permits, and stop matching a command name as a substring (#69)

### LINKED ISSUE #69: Convention hooks match a command name as a substring: ordinary greps are refused, and `rm -rf docs/dev-log` is not
## Parent

None — and that is part of the finding.

This was first reported as a straggler of #36: "four hooks never moved to the
shared tokeniser". That framing is wrong and should not be inherited. #36 is the
agent boundary — a pull request's base across four spellings, `worktree.baseRef`,
CLAUDE.md's false "hooks alone" claim, the unguarded `.claude/`, the stopping
rule, `probe`→`check`. Its hooks are `no-git-push.sh`, `no-pr-decisions.sh`,
`no-commit-to-main.sh` and `no-work-on-stale-branch.sh`.

The four hooks here — `pytest-via-uv-group.sh`, `alembic-via-uv-group.sh`,
`append-only-docs.sh`, `append-only-docs-edit.sh` — enforce CLAUDE.md
*conventions*: dependency groups and append-only history. They guard nothing an
agent could use to cross the boundary, and they were never enrolled in the #43
or #44 migrations. "Never moved to the shared tokeniser" implies a migration
they were never in, and it imports an urgency that belongs to the boundary
hooks. Filed separately for that reason.

It also turned out that the tokeniser is the fix for only half of what is
actually wrong here. See **Solution**.

## Context

Found by review on `dev-05` at `7cb4891`, and every verdict below was measured
against the hooks as they stand on that commit. `0` = allow, `2` = block.

The original report named the refusing direction only. The permitting direction
was then swept, and that is where the serious defects were.

### 1. A command name is matched as a substring, not at a command position

`pytest-via-uv-group.sh:12` and `alembic-via-uv-group.sh:12`:

```sh
if echo "$COMMAND" | grep -qE '(^|[;&|]|\s)(pytest|python[0-9.]*\s+-m\s+pytest)(\s|$)'; then
if echo "$COMMAND" | grep -qE '(^|[;&|]|\s)alembic(\s|$)'; then
```

Neither hook sources `lib/command-scan.sh`, so neither knows where a command
word is. A name appearing as an *argument* or as *text* satisfies the pattern,
and the allowlist that follows then fails to rescue it because prose does not
carry `uv run --group …`.

```
2  pytest-via-uv-group.sh   grep -rn pytest docs/
2  pytest-via-uv-group.sh   ls tests/ | grep pytest
2  pytest-via-uv-group.sh   git log --grep pytest
2  alembic-via-uv-group.sh  grep -rn alembic docs/
2  alembic-via-uv-group.sh  git log --grep alembic
2  alembic-via-uv-group.sh  echo "we run alembic upgrade head via the group"
```

Controls still behave:

```
2  pytest-via-uv-group.sh   pytest tests/
0  pytest-via-uv-group.sh   uv run --group test pytest tests/
2  alembic-via-uv-group.sh  alembic upgrade head
0  alembic-via-uv-group.sh  uv run --group migrations alembic upgrade head
```

**Why it is intermittent, which is worth recording.** The match needs a
separator or whitespace in front of the name, so whether a quoted mention is
refused depends on where in the quoted string it happens to fall:

```
0  pytest-via-uv-group.sh   echo "pytest lives in the test group"
2  alembic-via-uv-group.sh  echo "we run alembic upgrade head via the group"
```

The first is saved only because `"` sits immediately before the name and `"` is
not in the class. Same sentence shape, opposite verdict, on a difference that
has nothing to do with what either command would run.

### 2. `append-only-docs.sh` misses the whole-directory case

`append-only-docs.sh:17` requires a trailing slash:

```sh
APPEND_ONLY='docs/(dev-log|lessons-learned|eval-reports)/'
```

So the outer guard at line 19 never fires on the directory itself, and the
removal that destroys the most history is the one that passes:

```
0  append-only-docs.sh  rm -rf docs/dev-log
0  append-only-docs.sh  rm -rf docs/lessons-learned
0  append-only-docs.sh  rm -rf docs/eval-reports
0  append-only-docs.sh  mv docs/dev-log docs/archive
2  append-only-docs.sh  rm -rf docs/dev-log/      <- control, one character apart
2  append-only-docs.sh  rm docs/dev-log/x.md      <- control
```

A single trailing slash separates "every dev-log entry gone" from "blocked".

Two further routes are not covered by the `rm|mv|cp` verb list or the redirect
rule, both of which overwrite an existing entry in place:

```
0  append-only-docs.sh  truncate -s 0 docs/dev-log/x.md
0  append-only-docs.sh  tee docs/dev-log/x.md < new.md
2  append-only-docs.sh  : > docs/dev-log/x.md     <- control
```

### 3. `append-only-docs-edit.sh` compares an un-normalised path

The hook strips the project root by string prefix and then anchors `^docs/`. Any
spelling of the same file that does not reduce to that literal prefix is
allowed, on a file that exists:

```
2  docs/dev-log/devlog_2026-08-01_session-1.md            <- control
2  /abs/.../docs/dev-log/devlog_2026-08-01_session-1.md   <- control
0  ./docs/dev-log/devlog_2026-08-01_session-1.md
0  docs/../docs/dev-log/devlog_2026-08-01_session-1.md
```

A leading `./` is not an evasion — it is an ordinary spelling of a relative
path, and it is the shape of an ordinary mistake, which is what these hooks
exist to stop.

## Direction, and why that changes the priority

Defect 1 fails by **refusing**. That is the direction `lib/command-scan.sh`
chooses deliberately and says so in its header: over-splitting "can only refuse
more; it never hides one". A blocked `grep` is loud, visible, and one edit away.

Defects 2 and 3 fail by **permitting**, silently, on the append-only guarantee —
the same failure shape that justified `lib/command-scan.sh` in the first place.
They were not in the original report. They should be fixed first.

## Solution

Three fixes, not one. The unifying frame in the original report — "adopt the
shared tokeniser" — is correct for defect 1 and does nothing for defects 2 and 3,
which are a path-pattern bug and a path-normalisation bug respectively. Worth
stating, because "move these four onto `command-scan.sh`" would close the least
serious of the three and read as done.

**Stage 1 — the silent permits.**
- `append-only-docs.sh`: match the directory with or without its trailing slash,
  and extend the verb list to the in-place overwrite routes measured above.
- `append-only-docs-edit.sh`: resolve the path before comparing it, rather than
  stripping a literal prefix.

**Stage 2 — the substring match.** Route `pytest-via-uv-group.sh` and
`alembic-via-uv-group.sh` through `cs_normalise | cs_split` and anchor the name
at `^`. Checked against the cases above: `cs_split` yields `grep -rn pytest
docs/` as a single fragment whose command word is `grep`, so all six false
positives drop, and all six controls keep their verdicts (including `cd x && alembic upgrade head` and `python -m pytest tests/`, which stay blocked).

**Stage 3 — the check suite.** Every verdict in this issue belongs in it, in
both directions. The permitting cases matter more than the refusing ones and
none of them were pinned.

## Note on scope

`report-stale-branches.sh` also does not source `lib/command-scan.sh`. It reads
no command — it is a reporter, not a guard — so it is correctly absent from this
issue and from #36. Recording it here so the next sweep does not re-raise it.

bgunyel:
Fixes the three defects in issue #69, in the order the issue sets: the silent
permits first, the substring match second, the checks alongside each.

The four hooks here enforce CLAUDE.md *conventions* — dependency groups and
append-only history — rather than the agent boundary. Three of the four had no
coverage of any kind before this.

### Stage 1 — the silent permits (76a3eb8)

`append-only-docs.sh` matched the guarded directories only with a trailing
slash, so the outer guard never fired on the directory itself:

    ALLOW  rm -rf docs/dev-log
    BLOCK  rm -rf docs/dev-log/

Same command, one character apart, opposite verdicts, and the permitted half
takes every entry with it. The boundary is written out rather than made
optional, so `docs/dev-logbook` and `docs/dev-log.bak` stay other paths.
`truncate` and `tee` join the verb list: both overwrite an entry in place while
naming no redirect, so neither of the other rules saw them.

`append-only-docs-edit.sh` stripped the project root by string prefix and
anchored the remainder at `^docs/`, so it compared spellings rather than paths.
`./docs/dev-log/<entry>` and `docs/../docs/dev-log/<entry>` were permitted on a
file that exists. Both sides are normalised lexically now, by a function that
touches no filesystem, so the two cannot disagree.

### Stage 2 — the substring match (193510c)

These were the last two hooks reading a command without sourcing
`lib/command-scan.sh`, so each matched its name as an argument and as prose —
ordinary `grep`s and `git log`s refused. Intermittently: the match needed a
separator before the name, so a quoted mention was permitted or refused
according to whether a quote or a space happened to precede it. Both files
route through `cs_normalise | cs_split` and anchor at `^`.

The tokeniser's own header claimed to name every hook that reads a command, and
#63 had already found it stale once. It is checked now rather than maintained:
the suite reads which files source the library and asserts the paragraph names
each, in both directions.

### The review round (18bf1d8)

Five findings, four of them permitting and all four introduced by the two
commits above.

- Anchoring at `^` narrowed the guard to one runner: `poetry run pytest`,
  `uvx pytest`, `hatch run pytest`, `uv tool run pytest` and
  `poetry run alembic upgrade head` were BLOCK before the migration and ALLOW
  after it.
- `--group` was matched anywhere in the fragment, so the tool's own argument
  rescued the command — `uv run pytest --group test` was permitted while the
  comment beside the rule claimed otherwise.
- The intermittency the migration was meant to end survived one rule further
  down, in the file that had just fixed it.
- The fail-closed guard tested `cs_split` only, while both hooks also call
  `cs_normalise`; a library missing just that one permitted a bare `pytest`.
- Stage 1 normalised the Edit side and left the Bash side comparing spellings:
  `rm -rf docs//dev-log` and `rm -rf docs/./dev-log` were permitted.

### The review's finding (4d4b73d)

The runner list stopped one family early. `pipx run`, `micromamba run` and
`pixi run` were BLOCK on `dev-05` by the substring match and ALLOW once the rule
became a list, each being the sibling of something already on it. Now listed,
with six checks.

`xvfb-run` and `watch` reach pytest too and are deliberately **not** fixed here:
they take no subcommand and simply run the words after them, which is what
`cs_split` calls a wrapper word and already strips for `time`. They belong in
that list, which is #79 and in flight. Both are pinned as permitted, with
`time pytest tests/` pinned beside them as the contrast — when #79 adds them,
those two checks flip to BLOCK, which is intended rather than a regression.

Two silent permits on `dev-05` that the sections above did not claim are also
pinned now: the old allowlist was matched against the whole string, so
`cd tests && pytest` and `uv run --group test pytest && pytest tests/` were both
permitted.

### Evidence

`bash .claude/hooks/check-hooks.sh` — ALL CHECKS PASSED, 830 checks. Every
verdict the issue measured is pinned, in both directions, plus the review
round's.

Each fix was mutation-checked separately, and each turns exactly its own group
red: the directory boundary → the four whole-directory cases; the verb list →
the three in-place ones; the path normalisation → the three spellings; reading
the raw command → the three wrapper and control-word cases; the substring match
→ ten; the `uv run` rule → three; the runner list → eleven; the group prefix →
two; the word boundary → the one half of the pair that was permitted; the
leading-slash widening → four; guarding `cs_split` alone → two; dropping the two
new names from the tokeniser header → those two.

`make test`: 591 passed, 5 xfailed. `make audit`: no issues. No dependency
changed on this branch.

### Deliberately not done

`report-stale-branches.sh` reads no command — it is a reporter, not a guard —
and the issue records it as correctly out of scope.

Three trades are taken knowingly and each is a check rather than a silence:
`tee -a` appends and is refused with the rest of `tee`; a quote is a word
boundary, so `uv run echo "pytest …"` is refused rather than permitting
`uv run "pytest"`; and these two hooks carry no wrapper rule, so
`bash -c "pytest tests/"` is not read — as it was not before.

https://claude.ai/code/session_01TJLh3YRWhAAE98qMrtMmQ5

### DIFF
--- a/.claude/hooks/alembic-via-uv-group.sh
+++ b/.claude/hooks/alembic-via-uv-group.sh
@@ -47,7 +47,7 @@ fi
 # sanctioned one and is asked for the group; the rest cannot name a dependency
 # group at all, so naming alembic is enough to refuse.
 UV_RUN='^uv[[:space:]]+run([[:space:]]|$)'
-OTHER_RUNNER='^(uvx|uv[[:space:]]+tool[[:space:]]+run|poetry[[:space:]]+run|pdm[[:space:]]+run|hatch[[:space:]]+run|pipenv[[:space:]]+run|rye[[:space:]]+run|conda[[:space:]]+run|nix[[:space:]]+run)([[:space:]]|$)'
+OTHER_RUNNER='^(uvx|uv[[:space:]]+tool[[:space:]]+run|poetry[[:space:]]+run|pdm[[:space:]]+run|hatch[[:space:]]+run|pipenv[[:space:]]+run|rye[[:space:]]+run|conda[[:space:]]+run|micromamba[[:space:]]+run|pixi[[:space:]]+run|pipx[[:space:]]+run|nix[[:space:]]+run)([[:space:]]|$)'
 # A quote ends the name as a space does. See the trade in the companion file.
 NAME='(^|[^A-Za-z0-9_.-])alembic([^A-Za-z0-9_.-]|$)'
 
--- a/.claude/hooks/alembic-via-uv-group.sh
+++ b/.claude/hooks/alembic-via-uv-group.sh
@@ -13,14 +13,22 @@
 # `grep -rn alembic docs/`, `git log --grep alembic` and
 # `echo "we run alembic upgrade head via the group"` were all refused, and the
 # allowlist beneath could not rescue them because prose does not carry
-# `uv run --group migrations`. The companion file carries the fuller note,
-# including why the verdict on a quoted mention depended on which character
-# happened to precede the name; the fix is the same one and is made the same
-# way, because a rule answered twice is how these two came to disagree with the
-# four hooks that already knew the answer.
+# `uv run --group migrations`.
+#
+# The companion file carries the fuller note: why the verdict on a quoted
+# mention depended on which character happened to precede the name, why the
+# rule about runners is a list, why the group must be named before the tool,
+# and what a quote costs as a word boundary. The fix is the same one and is
+# made the same way, deliberately -- a rule answered twice is how these two
+# came to disagree with the four hooks that already knew the answer, and the
+# only honest way to keep them in step is to keep them identical.
 LIB="$(dirname "$0")/lib/command-scan.sh"
 [ -r "$LIB" ] && . "$LIB"
-if ! command -v cs_split >/dev/null 2>&1; then
+# Both functions this file calls, not just the one that names the file's
+# subject. Testing cs_split alone left cs_normalise unguarded, and a library
+# missing only that one permitted a bare `alembic upgrade head` silently.
+if ! command -v cs_split >/dev/null 2>&1 \
+   || ! command -v cs_normalise >/dev/null 2>&1; then
   echo "Blocked: alembic-via-uv-group.sh could not load lib/command-scan.sh, so it cannot tell an alembic invocation from a mention of one. Refusing rather than permitting." >&2
   exit 2
 fi
@@ -35,15 +43,30 @@ if printf '%s\n' "$CMDS" | grep -qE '^alembic([[:space:]]|$)'; then
   exit 2
 fi
 
-# `uv run` that reaches alembic without naming the group. Scoped to the one
-# fragment, so the group has to be named by the command that runs alembic and
-# not merely somewhere on the line, which is what the allowlist asked.
-if printf '%s\n' "$CMDS" \
-   | grep -E '^uv[[:space:]]+run([[:space:]]|$)' \
-   | grep -E '(^|[[:space:]])alembic([[:space:]]|$)' \
-   | grep -qvE -e '--group[[:space:]]+migrations([[:space:]]|$)'; then
-  echo "Blocked: uv run reaches alembic without the 'migrations' dependency group. CLAUDE.md pins Alembic to that group. Run it as: uv run --group migrations alembic <args>" >&2
-  exit 2
-fi
+# A runner standing where a command word goes, naming alembic. `uv run` is the
+# sanctioned one and is asked for the group; the rest cannot name a dependency
+# group at all, so naming alembic is enough to refuse.
+UV_RUN='^uv[[:space:]]+run([[:space:]]|$)'
+OTHER_RUNNER='^(uvx|uv[[:space:]]+tool[[:space:]]+run|poetry[[:space:]]+run|pdm[[:space:]]+run|hatch[[:space:]]+run|pipenv[[:space:]]+run|rye[[:space:]]+run|conda[[:space:]]+run|nix[[:space:]]+run)([[:space:]]|$)'
+# A quote ends the name as a space does. See the trade in the companion file.
+NAME='(^|[^A-Za-z0-9_.-])alembic([^A-Za-z0-9_.-]|$)'
+
+while IFS= read -r FRAGMENT; do
+  echo "$FRAGMENT" | grep -qE "$NAME" || continue
+  if echo "$FRAGMENT" | grep -qE "$OTHER_RUNNER"; then
+    echo "Blocked: that runner reaches alembic outside the 'migrations' dependency group, which it cannot name. CLAUDE.md pins Alembic to that group. Run it as: uv run --group migrations alembic <args>" >&2
+    exit 2
+  fi
+  echo "$FRAGMENT" | grep -qE "$UV_RUN" || continue
+  # Everything before the name. The group has to be an argument of uv, not of
+  # alembic, and this is what tells the two apart.
+  if ! printf '%s' "${FRAGMENT%%alembic*}" \
+       | grep -qE -e '--group[[:space:]]+migrations([[:space:]]|$)'; then
+    echo "Blocked: uv run reaches alembic without the 'migrations' dependency group named first. CLAUDE.md pins Alembic to that group. Run it as: uv run --group migrations alembic <args>" >&2
+    exit 2
+  fi
+done <<CS_FRAGMENTS
+$CMDS
+CS_FRAGMENTS
 
 exit 0
--- a/.claude/hooks/alembic-via-uv-group.sh
+++ b/.claude/hooks/alembic-via-uv-group.sh
@@ -6,14 +6,44 @@
 # A bare `alembic ...` runs outside that group and resolves its environment
 # differently, which is the failure mode the doc warns about. Every sanctioned
 # invocation in CLAUDE.md and pyproject.toml carries `uv run --group migrations`.
+#
+# Issue #69: this hook and pytest-via-uv-group.sh were the two that read a
+# command without knowing where one starts. The name was matched after a
+# separator or any whitespace, so it matched as an ARGUMENT and as PROSE --
+# `grep -rn alembic docs/`, `git log --grep alembic` and
+# `echo "we run alembic upgrade head via the group"` were all refused, and the
+# allowlist beneath could not rescue them because prose does not carry
+# `uv run --group migrations`. The companion file carries the fuller note,
+# including why the verdict on a quoted mention depended on which character
+# happened to precede the name; the fix is the same one and is made the same
+# way, because a rule answered twice is how these two came to disagree with the
+# four hooks that already knew the answer.
+LIB="$(dirname "$0")/lib/command-scan.sh"
+[ -r "$LIB" ] && . "$LIB"
+if ! command -v cs_split >/dev/null 2>&1; then
+  echo "Blocked: alembic-via-uv-group.sh could not load lib/command-scan.sh, so it cannot tell an alembic invocation from a mention of one. Refusing rather than permitting." >&2
+  exit 2
+fi
+
 INPUT=$(cat)
 COMMAND=$(echo "$INPUT" | jq -r '.tool_input.command')
+CMDS=$(printf '%s\n' "$COMMAND" | cs_normalise | cs_split)
 
-if echo "$COMMAND" | grep -qE '(^|[;&|]|\s)alembic(\s|$)'; then
-  if ! echo "$COMMAND" | grep -qE 'uv\s+run\s+.*--group\s+migrations'; then
-    echo "Blocked: bare alembic invocation. CLAUDE.md pins Alembic to the 'migrations' dependency group. Run it as: uv run --group migrations alembic <args>" >&2
-    exit 2
-  fi
+# alembic standing where a command word goes.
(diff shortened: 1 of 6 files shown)


bgunyel:
## Review: acceptance criteria in #69 are met

Verified by running the four hooks against every verdict the issue measured, in
both directions, rather than by reading the description — **41/41 match the
required outcome**. `0` = allow, `2` = block, as in the issue.

### Stage 1 — the silent permits

Closed, including the neighbours that had to stay untouched:

```
2  rm -rf docs/dev-log            2  mv docs/dev-log docs/archive
2  rm -rf docs/lessons-learned    2  truncate -s 0 docs/dev-log/x.md
2  rm -rf docs/eval-reports       2  tee docs/dev-log/x.md < new.md
0  rm -rf docs/dev-logbook        0  rm -rf docs/dev-log.bak
0  cat docs/dev-log/x.md          0  echo hi >> docs/dev-log/x.md
```

The Edit side normalises rather than compares spellings: `./docs/dev-log/<entry>`
and `docs/../docs/dev-log/<entry>` both refuse now, the absolute and bare
spellings keep their verdicts, and the two behaviours the directory exists for
are intact — a **new** entry file is permitted, an existing one is not, and
`README.md` stays revisable.

### Stage 2 — the substring match

All six false refusals drop, and every control holds:

```
0  grep -rn pytest docs/          2  pytest tests/
0  ls tests/ | grep pytest        2  python -m pytest tests/
0  git log --grep alembic         2  cd x && alembic upgrade head
0  echo "pytest lives in ..."     0  uv run --group test pytest tests/
```

### Stage 3 — the check suite

Real rather than decorative: **818 checks, 0 failures**, and I grepped the suite
for each of the issue's literal cases individually — the Edit-side spellings
included — to confirm they are pinned rather than merely passing. The bidirectional
check on the tokeniser header is the right answer to a comment that had gone
stale twice.

### Claims checked independently

`make audit` clean; no dependency or Makefile change on the branch. `make test`
reported 590 passed / 5 xfailed / **1 failed** for me — but
`test_installed_packages_match_uv_lock` fails identically on the `dev-05`
baseline, so it is venv drift in my throwaway worktree, not this branch.

### The PR under-claims two fixes

The old allowlist was matched against the whole string, so any sanctioned
invocation rescued a bare one elsewhere in the command. Both of these were
silent permits on `dev-05` and block now, and neither is mentioned:

```
dev-05: 0  ->  PR: 2    (cd tests && pytest)
dev-05: 0  ->  PR: 2    uv run --group test pytest && pytest tests/
```

---

## One finding: the runner list omits `pipx run`

Anchoring at `^` traded a substring match for a list, and the list is one word
short. The old hook refused `<anything> pytest`; the new one refuses only a
listed runner. Measured, `dev-05` → this branch:

```
2 -> 0    pipx run pytest
2 -> 0    pipx run alembic upgrade head
2 -> 0    micromamba run pytest
2 -> 0    pixi run pytest
2 -> 0    xvfb-run pytest tests/
2 -> 0    watch pytest
```

The PR declares the general shape — *"a runner not on it is permitted"* — so this
is not undisclosed. But `pipx run` is not a considered boundary. It is a
mainstream sibling of `uvx` and `uv tool run`, **both of which are on the list**,
and it reaches pytest outside the group in exactly the way the listed ones do.
This is the same class 18bf1d8 already caught once (*"anchoring at `^` narrowed
the guard to one runner"*); the list was extended then and stopped one name early.
It fails in the permitting direction, which the issue argues is the serious one.

Low severity in practice — nobody reaches for `pipx run pytest` in a uv/make
repo — and it is a one-word fix in both hooks plus a pinned check either way.

Distinct from #79, which is about the *wrapper-word* list (`sudo`, `env`,
`xargs`, `time`, `nice`, `nohup`) and `sh -c`. Those all still refuse correctly
here, because `cs_split` strips them; the six above are permitted precisely
because it does not.

The other three trades — `tee -a` refused despite appending, `bash -c "pytest …"`
not read, `find docs/dev-log -delete` outside the verb list — are each stated in
the PR and each within what the issue scoped.

**Recommendation: approve.** The `pipx run` gap is worth a follow-up issue rather
than a block on this one.

https://claude.ai/code/session_01D9c1CTaD3v6DhKL9AAKXSf

bgunyel:
Taken, and fixed rather than deferred — it is a one-word change per family plus
the checks, and it fails in the permitting direction. `4d4b73d`.

**Three of the six are now on the list.** `pipx run`, `micromamba run` and
`pixi run` — each the sibling of something already there (pipx beside `uvx` and
`uv tool run`, micromamba beside `conda run`, pixi beside `poetry run`). Six
checks, three per hook, and removing the three names fails exactly those six.

You are right that this is the same class `18bf1d8` already caught once. It is
recorded that way in the hook and in the commit message rather than as a fresh
discovery: narrowing a substring match to a list costs whatever the list omits,
and what it omits gets found by someone asking, not by the rule.

**The other two are deliberately not fixed here, and I think the distinction is
load-bearing rather than pedantic.** `xvfb-run` and `watch` take no subcommand;
they just run the words after them. That is what `cs_split` calls a wrapper
word, and it already strips `time`, `sudo`, `nice` and the rest that way.
Measured:

```
time pytest tests/       -> cs_split -> "pytest tests/"      (BLOCK, no list involved)
xvfb-run pytest tests/   -> cs_split -> "xvfb-run pytest tests/"
```

So they are a wrapper question, not a runner question. Putting them in the
runner list would answer "what is a wrapper word" in a third place — the habit
`lib/command-scan.sh` exists to end — and it would fix these two hooks while
leaving the four boundary hooks just as blind to `xvfb-run git push --all
origin`. They belong in `cs_split`'s list, which is #79, and I have not touched
that file because #79 is in flight.

They are pinned as ALLOW rather than left unsaid, with `time pytest tests/`
pinned beside them as the contrast. **When #79 adds them those two checks flip
to BLOCK — that is the intended outcome, not a regression**, and the comment at
both sites says so, so whoever lands #79 changes them there rather than
wondering.

**On the two fixes the PR under-claimed** — thank you, both are pinned now. The
old allowlist was matched against the whole string, so one sanctioned
invocation rescued a bare one beside it:

```
dev-05: 0  ->  2    cd tests && pytest
dev-05: 0  ->  2    uv run --group test pytest && pytest tests/
```

plus the alembic spelling of the second. `cs_split` judges each command on its
own, so neither rescue survives. The PR body is updated.

**On `make test`** — your reading is right and it is not this branch.
`test_installed_packages_match_uv_lock` failed for me too on the first run, in
this worktree, after merging the `dev-05` commits that moved `uv.lock`;
`uv sync --all-groups` installed `alembic` and `mako` and it has been 591/5
since. Worth saying plainly that the 590/1-failed number came from the venv,
not the lock.

`bash .claude/hooks/check-hooks.sh`: ALL CHECKS PASSED, 830 checks.
`make test`: 591 passed, 5 xfailed.

https://claude.ai/code/session_01TJLh3YRWhAAE98qMrtMmQ5

Issue #84, opened 2026-09-11
The library-load guard is missing in two hooks and too narrow in a third: a renamed cs_* silently permits force-push, PR merge and a push to main


bgunyel:
## Source

Two-axis code review of `dev-05` against `main` (2026-09-11), Standards axis.
Every verdict below was re-measured against the hooks as they stand on `dev-05`
at `c06225e`, by piping the command into the hook as the harness does, with the
hook copied into a fixture directory. `0` = allow, `2` = block.

## What is wrong

The four boundary hooks answer the question *"what do I do when
`lib/command-scan.sh` is not loadable?"* in three different ways, and two of
those answers permit. `CLAUDE.md:159-167`:

> **Guard code is a third standard of care.** … it can permit an act that
> closes every open pull request. … The shared tokeniser alone has had nine
> defects found by review rather than by the suite … Every one was silent and
> in the permitting direction.

This is a tenth, in the same direction, in the load of that tokeniser rather
than in its body.

`no-commit-to-main.sh:61-66` and `no-work-on-stale-branch.sh:263-275` test the
file before sourcing and the functions after. `no-git-push.sh:50` and
`no-pr-decisions.sh:86` do neither:

```sh
. "$(dirname "$0")/lib/command-scan.sh"
```

With no `set -e`, an unreadable or incomplete library leaves `cs_*` undefined.
`RAW=$(cs_git_args "$VERB") || continue` cannot tell *"not this verb"* from
*"no such function"*, so every verb falls through and the hook `exit 0`s.

### Measured: library absent

| command | `no-git-push` | `no-pr-decisions` | `no-commit-to-main` | `no-work-on-stale-branch` |
|---|---|---|---|---|
| `git push --force origin dev-05` | **allow** | — | — | block |
| `gh pr merge 81` | — | **allow** | — | — |
| `git push origin HEAD:main` | — | — | block | — |

### Measured: library present, `cs_git_args` / `cs_gh_args` renamed away

This is the likelier failure in practice — a rename during a refactor, not a
deleted file — and it takes a third hook with it:

| command | verdict |
|---|---|
| `git push --force origin dev-05` (`no-git-push.sh`) | **allow** |
| `gh pr merge 81` (`no-pr-decisions.sh`) | **allow** |
| `gh pr create --base main` (`no-pr-decisions.sh`) | **allow** |
| `git push origin HEAD:main` (`no-commit-to-main.sh`) | **allow** |
[...]

PR #89, opened 2026-09-11
Teach the wrapper rules the prefix words cs_split already strips

### LINKED ISSUE #79: The wrapper rules do not know the prefix words cs_split already strips: sudo sh -c evades every hook

bgunyel:
Closes #79. (The keyword does not link on a `dev-NN` base — the Development
section stays empty, so the issue needs closing by hand on merge.)

## What was wrong

`cs_split` has stripped the words that run another command since the fourth
review. The four wrapper regexes in the hooks never consulted that list, so one
word was recognised in one half of `lib/command-scan.sh` and invisible in the
other:

```
BLOCK   sudo git push --all origin            the push is at a command position
ALLOW   sudo sh -c 'git push --all origin'    the wrapper is not at ^
```

The same held for `timeout`, `xargs`, `nohup` and `env`, against all four
hooks. One question, two places, two answers — the defect class the header of
that file opens by naming.

## What changed

- The list is a variable. `cs_split` reads it through `awk -v`; `CS_WRAPPER_RE`
  interpolates it. One copy, not two.
- `CS_WRAPPER_RE` itself is derived once and grepped by all four hooks, which
  each carried their own copy of it before.
- The anchor admits a prefix word with its options and up to three further
  tokens — the bound `cs_split` already offers its tail to.

## The trade, recorded

- **Named and not closed.** `python3 -c`, `perl -e`, `find -exec sh -c`, a make
  target, a script written and then run: all still ALLOW, each pinned as a
  check that *wants* ALLOW so the limit is recorded rather than discovered.
- **Widened, not dropped.** `grep -rn 'sh -c .*git push' .claude/hooks/` stays
  ALLOW — dropping the anchor would refuse it, and that is what a session
  working on these files writes.
- **The quote-blind anchor** stays a soft spot, named in the header with the
  decision rather than an excuse: asking `cs_split` *would* answer it, and what
  stops this rule asking is that the suite pins it as reading raw text. A
  candidate for its own ticket.
- **Cost of widening:** `sed -i 's|sudo sh -c git push|X|' f.sh` was ALLOW and
  is now BLOCK. Both spellings pinned.

## One defect this change would have introduced

Making the list a variable added a state that probing for the functions cannot
detect: every function present, list empty, `cs_split` stripping nothing,
`sudo git push --all origin` permitted again. The first answer covered two of
four hooks under a comment saying the state was handled. The fail-safe is in
the library now — an empty list gives an empty `CS_WRAPPER_RE`, which `grep`
matches against every line, so all four refuse. Found by review of the first
commit; the second commit is that answer.

## Verification

- `bash .claude/hooks/check-hooks.sh` — **ALL CHECKS PASSED**.
- **Nine mutants, no survivors**: anchor reverted, anchor dropped, either word
  list emptied, `cs_split` carrying the list as a literal again, the empty-list
  fail-safe removed, the intervening token narrowed, either hook's probe
  removed.
- Two checks exist because mutation found the first version of them silent.
- `make test`: 590 passed, 5 xfailed, 1 failed — `test_environment_sync.py`,
  pre-existing venv drift (`alembic`, `mako` from the `migrations` group), not
  touched by this branch, which is six shell files.
- `make upgrade-safe` not run: no dependency, lockfile or Python change here.

https://claude.ai/code/session_01PGuqGrYJD4tBbuafJD6tFP

### DIFF
--- a/.claude/hooks/check-hooks.sh
+++ b/.claude/hooks/check-hooks.sh
@@ -146,18 +146,47 @@ ON_DEV="$FIXTURES/on-dev"
 mkdir -p "$FIXTURES/nolib"
 cp no-commit-to-main.sh "$FIXTURES/nolib/"
 
-# And a copy beside a library that loads, defines every function, and holds no
+# And a hook beside a library that loads, defines every function, and holds no
 # wrapper words. Since #79 the list cs_split strips is a variable rather than a
 # literal in its own regex, so this is a state the file can be in that probing
-# for the functions does not detect -- and one in which it strips no prefix and
-# the anchor admits none, silently and in the permitting direction. The
-# assignments are emptied by appending to the real library rather than by
-# rewriting it, so the fixture cannot drift from what it is a copy of.
-mkdir -p "$FIXTURES/emptylist/lib"
-cp no-commit-to-main.sh "$FIXTURES/emptylist/"
-cp lib/command-scan.sh "$FIXTURES/emptylist/lib/"
-printf 'CS_WRAP_OPTION_WORDS=""\nCS_WRAP_OPERAND_WORDS=""\nCS_WRAP_WORDS=""\n' \
-  >> "$FIXTURES/emptylist/lib/command-scan.sh"
+# for the functions does not detect -- every function is there and the strip
+# simply does nothing, silently and in the permitting direction.
+#
+# Written once and called per hook. Two copies of this builder, differing only
+# in the directory and the hook, was Duplicated Code inside the one change whose
+# subject is not answering a question in two places; found by review, not here.
+#
+# The assignments are emptied IN PLACE, not appended to. CS_WRAPPER_RE is
+# derived from them when the file is sourced, so an append lands after that
+# derivation, leaves the anchor built from the full list, and the fixture would
+# then test cs_split's strip while claiming to test the library's own fail-safe.
+# The first version of these checks appended, and that is what it measured.
+#
+# The sed is verified rather than assumed, for the reason the half-library
+# fixture beside it gives: one that silently changed nothing would leave every
+# check against it green and proving nothing.
+emptied_lib_fixture() {  # emptied_lib_fixture <name> <hook>
+  local dir="$FIXTURES/$1"
+  mkdir -p "$dir/lib"
+  cp "$2" "$dir/"
+  sed -E 's/^CS_WRAP_OPTION_WORDS=.*/CS_WRAP_OPTION_WORDS=""/;
+          s/^CS_WRAP_OPERAND_WORDS=.*/CS_WRAP_OPERAND_WORDS=""/' \
+      lib/command-scan.sh > "$dir/lib/command-scan.sh"
+  grep -q '^CS_WRAP_OPTION_WORDS=""$' "$dir/lib/command-scan.sh" \
+    && grep -q '^CS_WRAP_OPERAND_WORDS=""$' "$dir/lib/command-scan.sh" || {
+    echo "the $1 fixture did not empty the word list; its checks prove nothing" >&2
+    exit 1
+  }
+}
+#
+# All four hooks get one, because the fail-safe for an empty list is in the
+# library and therefore answers for all four at once. A per-hook probe would
+# have covered the two that already guard their load and left the other two
+# permitting -- which is what the first version of this did.
+emptied_lib_fixture emptylist       no-commit-to-main.sh
+emptied_lib_fixture emptylist-push  no-git-push.sh
+emptied_lib_fixture emptylist-pr    no-pr-decisions.sh
+emptied_lib_fixture emptylist-stale no-work-on-stale-branch.sh
 
 # check, with the hook's working directory named rather than inherited. The
 # hook is invoked by absolute path because it sources lib/ relative to $0.
@@ -960,6 +989,19 @@ unarmed 'nor no-pr-decisions.sh' \
         no-pr-decisions.sh "grep -qE '$WRAPRE"
 unarmed 'nor no-work-on-stale-branch.sh' \
         no-work-on-stale-branch.sh "grep -qE '$WRAPRE"
+# In either spelling. The four pins above name the single-quoted one, which is
+# how all four hooks wrote it; a re-derivation reached for with double quotes
+# would satisfy every one of them and answer the list a second time anyway.
+# Found by review of this change: a pin on one spelling of a literal is
+# evidence about that spelling and about nothing else.
+unarmed 'no-git-push.sh does not carry it double-quoted either' \
+        no-git-push.sh "grep -qE \"$WRAPRE"
+unarmed 'nor no-commit-to-main.sh' \
+        no-commit-to-main.sh "grep -qE \"$WRAPRE"
+unarmed 'nor no-pr-decisions.sh' \
+        no-pr-decisions.sh "grep -qE \"$WRAPRE"
+unarmed 'nor no-work-on-stale-branch.sh' \
+        no-work-on-stale-branch.sh "grep -qE \"$WRAPRE"
 
 echo "=== REGRESSION: issue #79, the wrapper rules did not know the prefix words ==="
 # cs_split has always stripped the words that run another command -- sudo, env,
@@ -997,6 +1039,14 @@ check no-git-push.sh BLOCK 'timeout -s KILL 30 + wrapped'   "timeout -s KILL 30 
 # point -- they measure the number rather than a command.
 check no-git-push.sh BLOCK 'three tokens before the wrapper'     "sudo a b c sh -c 'git push --all origin'"
 check no-git-push.sh ALLOW 'and four is past where it looks'     "sudo a b c d sh -c 'git push --all origin'"
+# What that run admits, where cs_split's tail would stop. cs_split breaks its
+# tail at a token opening a quote; this does not, because nothing here reads a
+# token as a command -- they are skipped on the way to a wrapper word that must
+# still appear. The asymmetry is deliberate, argued at CS_WRAP_TOKEN, and in
+# the refusing direction, so it is pinned as a BLOCK rather than left to be
+# rediscovered as a divergence.
+check no-git-push.sh BLOCK 'a quoted token does not end the run' \
+         "sudo \"x\" sh -c 'git push --all origin'"
 # On a dev branch, so that the branch cannot be what answers for the wrapper
 # rule -- the same care the #68 wrapped-commit check takes above.
 check_in "$ON_DEV" no-commit-to-main.sh BLOCK 'sudo + wrapped commit' \
@@ -1009,9 +1059,17 @@ check_in "$ON_DEV" no-commit-to-main.sh BLOCK 'env + wrapped push' \
          "env FOO=1 sh -c 'git push --all origin'"
 check_in "$ON_DEV" no-commit-to-main.sh BLOCK 'nohup + wrapped push' \
          "nohup sh -c 'git push --all origin'"
-# The third and fourth consumers. A decision and a commit on a branch whose life
-# is over are the two the other hooks answer for, and both carried the same
-# blind spot.
+# The third consumer. A decision is what this hook answers for, and it carried
+# the same blind spot.
+#
+# The fourth is no-work-on-stale-branch.sh, and it is NOT here: its verdicts
+# need a worktree on a branch whose life is over, and those fixtures are built
+# further down. Its prefix-word checks are in that section, beside its other
+# wrapper ones. This comment said "the third and fourth consumers" with three
+# no-pr-decisions checks under it and nothing for the fourth anywhere -- found
+# by review of this change, which is the letter of "no fix lands without a
+# check that fails without the fix" going unmet while an armed pin on
+# CS_WRAPPER_RE carried the substance.
 check no-pr-decisions.sh BLOCK 'timeout + wrapped merge'    "timeout 5 sh -c 'gh pr merge 5'"
 check no-pr-decisions.sh BLOCK 'sudo + wrapped release'     "sudo bash -c 'gh release create v1'"
 check no-pr-decisions.sh BLOCK 'nohup + wrapped eval merge' "nohup eval 'gh pr merge 5'"
@@ -1055,11 +1113,15 @@ check no-git-push.sh ALLOW 'find -exec sh -c is out of reach' \
 echo "=== issue #79: the soft spot the anchor keeps, and what widening cost it ==="
 # The anchor carries its own separator class, and that class knows nothing about
 # quoting -- so a verdict still turns on a sed delimiter, which is the complaint
-# #68 was filed about. #68 fixed it in cs_split and cannot fix it here: the fix
-# is to ask cs_split, and this rule must read raw text because a wrapped payload
-# has no command position and `bash <<EOF` is itself a wrapper. Both spellings
-# are pinned side by side, because it is the delimiter that decides the verdict
-# and that is the part that reads as arbitrary from inside a session.
+# #68 was filed about. It is deferred rather than impossible, and the reason is
+# argued once, at CS_WRAPPER_RE in lib/command-scan.sh: asking cs_split WOULD
+# answer it, and what stops this rule asking is that the pins above assert it is
+# handed the raw command. This banner said "cannot fix it here" and gave the
+# raw-text reason, which is the version that header examines and rejects --
+# found by review of this change, one file asserting what the other disowns,
+# which is the shape #63 found in CLAUDE.md and the header found in itself.
+# Both spellings are pinned side by side, because it is the delimiter that
+# decides the verdict and that is the part that reads as arbitrary in session.
 check no-git-push.sh BLOCK 'the pipe delimiter satisfies the anchor' \
          "sed -i 's|sh -c git push --all|X|' f.sh"
 check no-git-push.sh ALLOW 'the slash delimiter does not' \
@@ -1098,6 +1160,17 @@ armed 'cs_split strips whatever that variable holds' \
       lib/command-scan.sh 'match(line, "^(" wrapwords ")[[:space:]]+")'
 armed 'and the anchor admits whatever the union holds' \
       lib/command-scan.sh '($CS_WRAP_WORDS)[[:space:]]+'
+armed 'the intervening token is named once and used once' \
+      lib/command-scan.sh '($CS_WRAP_TOKEN){0,3}'
+# The fail-safe, as a property of the file. It is what answers for the two
+# hooks that do not guard their own load, so a change that built the anchor
+# unconditionally would leave those two permitting and every behavioural check
+# above still green -- the emptylist ones are what go red, and they can only go
+# red while this branch exists.
+armed 'the full anchor is built only when both halves are present' \
+      lib/command-scan.sh '[ -n "$CS_WRAP_OPTION_WORDS" ] && [ -n "$CS_WRAP_OPERAND_WORDS" ]'
+armed 'and an empty list answers with an empty anchor, which matches every line' \
+      lib/command-scan.sh 'CS_WRAPPER_RE=""'
 # The literal cs_split carried before #79. It is the second copy this fix
 # removes, and a re-derivation would restore it.
 unarmed 'cs_split no longer carries the list as a literal' \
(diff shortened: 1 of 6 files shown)


bgunyel:
## Review: acceptance criteria met, and one defect this change reproduces inside its own fix

All seven acceptance criteria in #79 are met. Each was verified against this
branch rather than read off the description: `check-hooks.sh` run, verdicts
compared base-vs-branch on every shape the issue names, and four mutants run
independently of the nine claimed here.

One finding, and it is the reason this is not a plain approval. It is the same
asymmetry #79 was filed about, one option deeper, in the permitting direction,
and it contradicts a paragraph this branch adds.

---

### The finding: `CS_WRAP_TOKEN` is *narrower* than `cs_split`'s tail token, not wider

`lib/command-scan.sh:469`. The anchor's grammar is `word -> options* ->
token{0,3}`, and `CS_WRAP_TOKEN` excludes a token beginning with `-`:

```
CS_WRAP_TOKEN="[^-[:space:]][^[:space:]]*[[:space:]]+"
```

`cs_split`'s tail loop admits any token:

```awk
if (!match(rest, /^[^[:space:]]+[[:space:]]+/)) break
```

So an option appearing **after** a separated option value stops the anchor
dead, while `cs_split` walks straight past it and finds the command. Measured
on this branch, `no-git-push.sh`:

```
BLOCK   sudo -u root -n git push --all origin
ALLOW   sudo -u root -n sh -c 'git push --all origin'

BLOCK   nice -n 10 -- git push --all origin
ALLOW   nice -n 10 -- sh -c 'git push --all origin'

BLOCK   timeout -s KILL 30 --preserve-status git push --all origin
ALLOW   timeout -s KILL 30 --preserve-status bash -c 'git push --all origin'
```

`cs_split` on the first of those yields the push at the third candidate:

```
root -n git push --all origin
-n git push --all origin
git push --all origin
push --all origin
```

One word recognised in one half of the library and invisible in the other,
which is the sentence the header of that file opens with. `sudo -n`,
`nice -n 10 --` and `timeout --preserve-status` are ordinary spellings; `--`
in particular defeats the anchor whenever it follows a separated option value,
while `sudo -- sh -c` is correctly BLOCK because nothing has consumed an
operand yet.

**Why this is a header defect and not only a gap.** The paragraph added at
`CS_WRAP_TOKEN` says:

> WIDER THAN cs_split's TAIL TOKEN, deliberately, and this is the one place the
> two answers differ on purpose.

and gives the quote rule as that one place, and then argues the difference is
safe because

> admitting a token too many can only refuse more, never hide a wrapper.

The leading-dash exclusion is a second difference, it runs the other way, and
it is not argued for anywhere in the file. The direction argument is the
argument for widening; it does not license this narrowing. The comment claims
the classes differ in one respect when they differ in two, which is the shape
the same paragraph says review of this change already caught once.

Nothing pins it: the suite pins `sudo -u root sh -c` BLOCK and never asks what
a *second* option behind the operand does.

**The fix, tested rather than proposed.** One character class:

```
-CS_WRAP_TOKEN="[^-[:space:]][^[:space:]]*[[:space:]]+"
+CS_WRAP_TOKEN="[^[:space:]]+[[:space:]]+"
```

With that, measured:

- all three shapes above turn BLOCK
- the bound is unchanged: `sudo a b c sh -c` BLOCK, `sudo a b c d sh -c` ALLOW
- the #68 shapes stay ALLOW: `grep -rn 'sh -c .*git push' .claude/hooks/`,
  `grep -rn 'eval .*git push' .claude/hooks/`, the `notes.md` echo
- `sudo "x" sh -c` stays BLOCK, so the quoted-token pin is unaffected
- `bash check-hooks.sh` -> **ALL CHECKS PASSED**, zero failures

It also collapses the divergence rather than widening past it: the token class
then matches `cs_split`'s exactly, and the shared claim becomes the class and
the count rather than the count alone. The header paragraph needs rewriting to
say that, and the three shapes need checks -- a fix landing without a check
that fails without it is the letter this branch already took trouble to meet
for the fourth consumer.

---

### The criteria, as measured

| # | Criterion | Verdict |
|---|---|---|
| 1 | Prefix set admitted without a second copy | met -- list written once at `command-scan.sh:364`; no hook carries its own anchor, in either quoting |
| 2 | `sudo` / `timeout 5` / `xargs` / `env` / `nohup` each BLOCK a wrapped push, each checked | met -- all five flip ALLOW to BLOCK |
| 3 | Unwrapped controls still BLOCK, the three #68 shapes still ALLOW | met -- 5/5 and 3/3 |
| 4 | Out-of-reach shapes named in the header, `python3 -c` the example | met -- line 409 |
| 5 | A decision recorded on the quote-blind anchor | met -- ONE SOFT SPOT, line 417 |
| 6 | Mutation-checked, both directions | met -- see below |
| 7 | `bash .claude/hooks/check-hooks.sh` passes | met |

Beyond the five words the criterion names, all fourteen in `CS_WRAP_WORDS` were
asked individually and all fourteen BLOCK a wrapped push, `timeout` and `flock`
with their operand included. So do the stacked forms `sudo env sh -c`,
`sudo timeout 5 sh -c`, `stdbuf -o0 sh -c`, `sudo -- sh -c`, `sudo zsh -c`, and
`sudo bash <<EOF` with a push in the body.

### Mutants run here, independently of the nine

| Mutant | Result |
|---|---|
| Anchor reverted to its pre-#79 form | 26 checks red, led by the six prefix-word BLOCKs |
| Anchor dropped outright | 14 red, including exactly the three #68 ALLOW checks the criterion names |
| Both word lists emptied | 169 red; all four hooks still refuse the verb each answers for, and `ls` / `gh issue list` stay ALLOW -- the fail-safe is scoped as described |
| Fail-safe removed *and* lists emptied | 75 red, and `sudo git push --all origin` is ALLOW again |

The last of those is worth recording as confirmation rather than as a finding:
the defect this PR says it introduced and then fixed is real, the fail-safe in
the library is what answers for it, and it is genuinely pinned. Moving it out of
the hooks and into the library was the right call -- a per-hook probe would have
left `no-git-push.sh` and `no-pr-decisions.sh` permitting, as the comment says.

### Over-blocking sweep

Twelve realistic commands were asked of both trees. Exactly one verdict changed:

```
ALLOW -> BLOCK   sed -i 's|sudo sh -c git push --all|X|' f.sh
```

which is the cost this branch names and pins on both sides. `time make test`,
`env | grep GIT`, `xargs -I{} echo {} < files.txt`, `timeout 30 pytest tests/`,
`sudo apt-get install git`, `nohup npm run dev &`, and a commit message quoting
`sudo sh -c ... git push` were all unchanged.

### Minor

- `find . -exec sh -c` stays ALLOW. #79 listed it among the measured evasions;
  this branch reclassifies it as out-of-reach on the grounds that `-exec` is not
  a prefix position, and pins it. That reading is right and it is recorded, so
  it is noted rather than raised.
- No hook sets `-u`, so an unset `CS_WRAPPER_RE` from a library that never loads
  degrades to the empty string and the vacuously-true conjunct, as the comment
  claims. Checked, because the claim depends on it.
- Branch is one commit behind `dev-05` (`a1e09f9`, a readme change), and still a
  draft.
- `make test` was not re-run. The failure named in the description is a Python
  environment-sync test and this branch is six shell files, so it is reported
  here as unverified rather than as confirmed.

### Recommendation

Land it after `CS_WRAP_TOKEN` is widened, its paragraph corrected to claim the
class as well as the count, and the three shapes pinned. Everything else in the
change holds up under mutation, and the fail-safe reasoning in particular is
stronger than the criterion asked for.

bgunyel:
Finding taken, and it was right. Fixed in 6a67326.

I reproduced it on the branch before changing anything rather than reading it
off the report — all three pairs, exactly as described:

```
BLOCK  ->  BLOCK   sudo -u root -n git push --all origin
ALLOW  ->  BLOCK   sudo -u root -n sh -c 'git push --all origin'
BLOCK  ->  BLOCK   nice -n 10 -- git push --all origin
ALLOW  ->  BLOCK   nice -n 10 -- sh -c 'git push --all origin'
BLOCK  ->  BLOCK   timeout -s KILL 30 --preserve-status git push --all origin
ALLOW  ->  BLOCK   timeout -s KILL 30 --preserve-status bash -c 'git push ...'
```

`CS_WRAP_TOKEN` is `[^[:space:]]+[[:space:]]+` as proposed. Bound unchanged
(`sudo a b c sh -c` BLOCK, `sudo a b c d sh -c` ALLOW), the three #68 shapes
still ALLOW, `sudo "x" sh -c` still BLOCK, `check-hooks.sh` green.

### One correction to the summary, because this file's premise is that claims are true

> the token class then matches `cs_split`'s exactly, and the shared claim
> becomes the class and the count rather than the count alone

The class does. The two are still not identical, and the quoted `match` line is
only the second half of `cs_split`'s tail loop — the first half breaks on a
quote:

```awk
for (k = 0; k < 3; k++) {
  if (rest ~ /^["]/ || rest ~ /^[\x27]/) break
  if (!match(rest, /^[^[:space:]]+[[:space:]]+/)) break
```

That stop is why `sudo "x" sh -c 'git push --all origin'` is BLOCK here and
yields no candidate there. So the header now claims **the class and the count,
with one remaining difference in the loop** — deliberate, refusing direction,
and pinned — rather than parity. Adopting the stronger sentence would have been
the third wrong version of that paragraph, which is the one thing it can least
afford: it has now been wrong twice, in both directions, and it says so.

### Checks

Seven. Each of the three shapes in **both** halves, because a single verdict
says nothing about which half moved, plus `sudo -- sh -c` as the control that
was never broken — it is what says the three are about the token class and not
about `--`.

### Mutants: ten, no survivors

The two new ones are the point:

| Mutant | Red |
|---|---|
| Token excludes a leading dash again | exactly the three **wrapped** checks; the unwrapped three stay green — the asymmetry itself, isolated |
| Token excludes a dash and a quote | those three plus the quoted-token pin |

### Also

- Merged `dev-05`, which had moved ten commits since the review (including
  #69's hook work). Clean; suite green after.
- `make test` re-run: 590 passed, 5 xfailed, 1 failed —
  `test_environment_sync.py`, the venv missing `alembic`/`mako` from the
  `migrations` group. Unrelated to this branch, and unchanged by it.
- Still a draft, deliberately: flipping it to ready is yours to give. Say the
  word and I will, or use the button.

https://claude.ai/code/session_01PGuqGrYJD4tBbuafJD6tFP

bgunyel:
## Not ready to merge: the fix holds, the base has moved under it

**`6a67326` re-verified.** `check-hooks.sh` passes (905 ok, 0 FAIL). The three
shapes and `sudo -u root --login sh -c` all BLOCK a wrapped push, the bound of
three holds, `sudo "x" sh -c` BLOCKs, and the #68 shapes ALLOW. Both new mutants
turn red exactly the checks the reply names. The correction to the review is
right: the class matches `cs_split`'s, the loop does not, and the header now says
so. No further findings against the change itself.

What stops the merge is `dev-05`. #93 (#84, *"Answer the library-load question
once, in every file that sources it"*) landed after that verification. It edits
all six files this branch edits, and it rebuilt the load guards this branch
extends.

---

### 1. It does not merge

Trial merge of `6a67326` into `dev-05` at `bcb7bf1`:

```
CONFLICT (content): .claude/hooks/no-commit-to-main.sh
CONFLICT (content): .claude/hooks/no-work-on-stale-branch.sh
CONFLICT (content): .claude/hooks/check-hooks.sh        (2 hunks)
```

GitHub reports mergeability as `UNKNOWN` and there is no CI, so nothing on the
PR page says this.

Both hook conflicts are the same collision: each side extends the same load
guard. `dev-05` probes all three functions, and this branch adds the two
word-list probes.

```
dev-05:   ! command -v cs_normalise || ! command -v cs_split || ! command -v cs_git_args
#89:      ! command -v cs_split || [ -z "$CS_WRAP_OPTION_WORDS" ] || [ -z "$CS_WRAP_OPERAND_WORDS" ]
```

The union is the textual answer. Whether the word-list probe belongs in #84's
shared load contract across all four hooks, rather than in two of them, is the
design question under it.

### 2. The natural resolution of the suite conflict deletes the fail-safe's checks, and stays green

#84 moved every nolib and halflib check out of its per-hook section and into
the load-contract section at the foot of `check-hooks.sh`. That section says
why: keeping the lesson in one file's section is what let three hooks get it
wrong.

This branch wrote its emptied-list checks *inside* those per-hook sections. In
the trial merge, **all nine sit inside the conflicting hunks and none outside
them:**

```
emptylist        no-commit-to-main.sh        BLOCK x2
emptylist-push   no-git-push.sh              BLOCK, ALLOW
emptylist-pr     no-pr-decisions.sh          BLOCK, ALLOW
emptylist-stale  no-work-on-stale-branch.sh  BLOCK x2, ALLOW
```

`dev-05`'s side of each hunk is a comment pointing at the new section, and it
is the side a resolver will reach for, because #84 deleted those sections on
purpose. Taking it deletes every check that pins the empty-list fail-safe.
Nothing turns red, because a deleted check cannot fail. That brings back the
state the second commit of this branch exists to close:
`sudo git push --all origin` permitted by a library that loads with an empty
list, and a comment saying it is handled.

The nine checks, and the `emptied_lib_fixture` builder they rest on, need to
move into the load-contract section. Dropping them is the failure mode.

### 3. Once merged, the tree would contradict itself about which hooks guard their load

This branch argues in four places that `no-git-push.sh` and `no-pr-decisions.sh`
source the library unguarded, and uses that as the reason the fail-safe lives in
the library:

```
lib/command-scan.sh:522   "unguarded, so a per-hook probe would have covered two of four"
check-hooks.sh:205        "the two that already guard their load and left the other two"
check-hooks.sh:1213       "hooks that do not guard their own load"
check-hooks.sh:2185       "The other two hooks source the library unguarded"
```

(line numbers on this branch)

On `dev-05`, #84 made that false and pinned the opposite:

```
check-hooks.sh:3830   unarmed 'no-git-push.sh does not source the library unguarded'
check-hooks.sh:3832   unarmed 'no-pr-decisions.sh does not source the library unguarded'
```

Three of the four sit outside the conflict hunks and would merge silently. The
fourth, `check-hooks.sh:2185`, is inside the first `check-hooks.sh` hunk beside
the checks it explains, so it goes or moves with them. That is
one file asserting what another pins false, which is the shape #63 found in
CLAUDE.md. The fail-safe's *verdicts* survive: with a probe that checks
functions and not the list, an emptied list still passes the guard and the
library answers. It is the stated *reason* that goes stale. The argument for a
library-level fail-safe needs restating on #84's footing, or retiring if the
word-list probe moves into the shared guard.

---

### Before this can merge

1. Merge `dev-05` into the branch. Resolve both load guards as a deliberate
   decision about where the word-list probe lives, not as a union by default.
2. Move the nine emptied-list checks and their fixture builder into #84's
   load-contract section.
3. Rewrite or retire the four "unguarded" claims.
4. Re-run `check-hooks.sh`, then re-run the mutant that empties the word list
   **and** removes the fail-safe. It must still turn red. A green suite alone
   cannot show the coverage survived a resolution whose failure mode is deleting
   checks.
5. Leaving draft is Bertan's call. On merge, #79 needs closing by hand: the
   closing keyword does not link on a `dev-NN` base.

The change itself does not need revisiting. What needs doing is carrying it
across #84 without losing its checks on the way.

bgunyel:
Merged `dev-05` across #84 in c137797. GitHub now reports the PR **MERGEABLE / CLEAN**, 0 behind. One commit on purpose: the natural resolution is green and wrong, and no point in this branch's history should be that tree.

Taking the five steps in order.

### 1. The load guards: the word-list probe lives in no guard

Both conflicting guards are `dev-05`'s, verbatim. That is the decision, not the union by default, and the reason is a measurement your review did not have.

The empty-`CS_WRAPPER_RE` fail-safe never covered all four boundary hooks' siblings. The anchor is read by the four boundary hooks. `pytest-via-uv-group.sh` and `alembic-via-uv-group.sh` call `cs_split` and never the anchor. So with the list emptied, `sudo pytest tests/` and `sudo alembic upgrade head` had no runner at `^` and were **permitted**, the whole time that fail-safe was described as the answer. That makes three versions of this guard, each fitted to part of the set its comment claimed:

- a probe in two hooks of four
- an anchor fail-safe covering four consumers of six
- and #84 then falsified the "two hooks are unguarded" premise it was argued from

So the state is reduced to one #84 already answers. The list's only reader is `cs_split`, and the library now withdraws it (`unset -f cs_split`) when either half of the list is empty. An incomplete list then looks exactly like a renamed function. Every consumer the list could mislead calls `cs_split`. #84's derivation forces each probe set to equal its call set, so all six already probe it and refuse by name. No guard changed, and the derivation is untouched. The empty-anchor branch is gone with its premise: two layered fail-safes would each let a mutant removing the other survive.

`unset -f` has to stand *after* the definition, or the definition restores it. That position is a check, not an argument (see step 4).

### 2. The nine checks: rebuilt in the load-contract section, not dropped

`dev-05`'s side of both hunks was taken, and the checks were rebuilt beside `mk_halflib`:

- **The fixture:** `mk_emptylist`, driven off `LIB_CONSUMERS`, so **six** consumers, not four. It follows the same discipline: path derived once, edit made in place and asserted in both directions, the library asserted to load with every other function, the target asserted executable.
- **Per consumer:** a command it has no opinion about (so a BLOCK can only be the guard), a command it refuses *only* through `cs_split`'s prefix strip, and the refusal naming the file.
- **The mechanism:** `cs_split` present intact, and absent with both halves empty *and with each half alone*. This is deliberately a check rather than a fixture guard, because a guard would exit the suite and report a removed mechanism as an aborted run.
- **Pins:** the withdrawal as `armed` text, and an `unarmed` pin per consumer that no guard learned the list.

34 checks; suite at 1009 ok.

### 3. The "unguarded" claims: all four retired, and a sweep found no fifth

`command-scan.sh:522` went with the replaced paragraph. `check-hooks.sh:205` went with the removed builder. `:2185` was inside a hunk. `:1213` was the pair of `armed` pins on the empty anchor; those now point at the load contract. The claim's history is kept *as history* in `THE WORD LIST IS PART OF THE LOAD`.

### 4. Mutants: eleven, none survive, all runs completed

| Mutant | Red |
|---|---|
| **Your step 4 as written:** list emptied **and** withdrawal removed | 69 |
| **Withdrawal removed, list intact** | 22, **all 22 inside the word-list block and none outside it**, verified by section |
| Withdrawal moved above `cs_split`'s definition | 21 |
| Withdrawal tests the union (`&&` for `\|\|`) | 3: both half-empty checks and the pin |
| A hook learns the list instead | 1 |
| Anchor reverted / dropped / literal restored / token dash / dash+quote | 30 / 12 / 4 / 3 / 4 |

The second row is the evidence the coverage survived the resolution. With the list intact every ordinary check stays green, so only the rebuilt fixtures can see the mutant, and exactly they go red. A green suite alone could not have shown that.

### 5. Draft and #79

Still a draft; that's yours. On merge, #79 needs closing by hand, since the keyword does not link on a `dev-NN` base.

`make test`: 595 passed, 5 xfailed, 1 failed. The failure is the pre-existing `test_environment_sync.py` venv drift (`alembic`/`mako`), not touched by this branch.

https://claude.ai/code/session_01PGuqGrYJD4tBbuafJD6tFP

Issue #104, opened 2026-09-13
Give every boundary requirement a permanent ID, tag every check, and fail on an uncovered requirement


bgunyel:
Part of #103. Blocked by #94 and #98.

## Problem

Nothing connects #36's requirements to the checks that are meant to verify them, so "is this requirement verified?" has no answer short of reading the whole suite, and a requirement no check names is invisible. See #103 for the audit.

## What this issue builds

The requirements file, the tagging convention, the coverage check and the `--matrix` output, and it **tags every existing check**. It adds **no behavioural checks**: the matrix it produces is the list of gaps that #105 fills.

## Decisions it enforces (numbers refer to #103)

**`.claude/hooks/requirements.md`** (Q6)
- A header stating the stability rules: IDs are never renumbered or reused; a requirement that stops being true is marked, never deleted; new requirements append. It also states the Q16 rule: a PR that fixes a hook defect appends its `GH-<n>` entry and tags the check that fails without the fix.
- One entry per requirement:
  - `id`: `US-n`, `FR-n` or `GH-<issue>[.m]` (Q7, Q13, Q14)
  - `text`
  - `from`: the issue or section it came from
  - `kind`: for `GH-` entries only; `defect-permitting`, `defect-refusing` or `doc-claim`
  - `status`: `active`, `retired: <reason>`, `superseded-by: <ID>` (Q10), or `drifted: <evidence>` (Q22)
  - `seam: none` plus `verify: runbook §n | review | tests/<file>` (Q17), when no executable seam exists
  - `direction: refuse-only | permit-only` plus a reason (Q15), when a requirement is one-sided
- **Sources:**
  - `US-1`…`US-32`: #36's user stories, **verbatim**. Superseded ones are marked, not rewritten (Q10).
  - `FR-n`: #36's Implementation Decisions (stages 0–3) and both Amendments, one FR per behaviour that can be tested on its own. Mark #36's "no new seam is introduced" and "CLAUDE.md's section is trimmed to the operative paragraph" as `drifted`, with evidence (the suite's file-reading seam around line 3006; the section is now 87 lines).
  - `GH-<n>[.m]`: #43, #44, #47, #50, #51, #58, #61, #62, #63, #68, #69, #70, #71, #72, #73, #79, #84, and every sub-issue of #36 filed on 2026-09-13 that has closed by the time this lands (#94 and #98 at least, since they block this one). #61 is `seam: none, verify: tests/<file>` (Q30).
[...]

Issue #103, opened 2026-09-13
Verify the agent boundary against a stable requirements matrix


bgunyel:
Traces: #36

## Problem

#36 specified the agent boundary in 32 user stories, a set of implementation decisions and two amendments, and all of its stage tickets have closed. The evidence that the boundary holds is `.claude/hooks/check-hooks.sh`: 499 `check` calls, green at `origin/dev-05` 750aace. Nothing connects the two. The checks are grouped by the review that found each defect, not by the requirement they establish, so "is user story 9 verified?" is answered by reading a 4,000-line file. And "the suite is green" says nothing about a requirement no check names. The suite's own header says this: "a check suite is evidence about the cases it names and about nothing else".

An audit on 2026-09-13 mapped every story, decision and amendment onto the existing checks. It found:

- **Gaps.** No check reads the refusal messages of `no-git-push.sh` or `no-pr-decisions.sh` (story 7). Most `gh issue` subcommands are unchecked (story 14). No check verifies that every registered hook is exercised. The eight `PreToolUse` timeouts are unpinned. Mutation-checking exists only as prose.
- **Live defects**, filed as sub-issues of #36:
  - #94: a push from a subdirectory of the main checkout is permitted
  - #95: every hook permits when `jq` is absent or its input is malformed
  - #96: a line long enough to outlast the 5 s timeout is permitted
  - #97: `gh release edit --draft=false` and `upload` are permitted
  - #98: the suite counts a crashed hook as ALLOW
  - #99: `baseRef: head` forks from a stale local ref
  - #100, #101, #102: documentation drift

## Goal

A **stable traceability matrix**: every requirement the boundary has accumulated has a permanent ID, and the check suite itself fails when an active requirement is not covered. Then close the gaps and add the kinds of check the suite has lacked.

## Work issues (sub-issues of this one)

1. #104: requirements file, tagging, coverage check, `--matrix`
2. #105: gap-fill
3. #106: invariance families
4. #107: mutation harness
5. #108: robustness, environment and fail-direction checks
6. #109: timing, refusal messages, configuration integrity, cross-hook consistency
7. #110: live acceptance runbook

Order and blockers are native dependency edges: W1 ← #94, #98; W2, W3, W4, W7 ← W1; W5 ← W1, #95; W6 ← W1, #96, #97.

## Decisions

Taken in a grilling session between Bertan and the assistant on 2026-09-13. Numbers are the session's question numbers, so that a child issue can cite one.

**What this work is**
[...]

Issue #98, opened 2026-09-13
check-hooks.sh reads any exit other than 2 as ALLOW: a crashed hook passes every permitting check


bgunyel:
Part of #36. Found by the requirements audit recorded in #103.

## Problem

`check-hooks.sh`'s seam-1 helpers read a hook's exit code as one bit: `if [ $rc -eq 2 ]; then got=BLOCK; else got=ALLOW; fi` (the `check` and `check_in` helpers at `origin/dev-05` 750aace; `flip` the same way). So **any** exit other than 2 counts as ALLOW: 1 from a syntax error, 127 from a missing interpreter or tool, 124 from `timeout`, or a crash partway through. A hook that is broken outright passes every ALLOW-expecting check in the suite.

The suite already knows the shape of this defect. The comment above its fixtures reads: "An unmade fixture would make `( cd "$dir" && hook )` return 1, which reads as ALLOW -- so every ALLOW-expecting check below would pass without running the hook at all." It guards that one cause, not the class.

Every committed hook exits only 0 or 2 today (audit, by reading and by running), so this finds no defect in the hooks. It is a defect in the evidence.

kind: `defect-permitting` (of the suite)

## Decisions (from the grilling recorded in #103)

- **Q20 — ALLOW means exit exactly 0.** Exit 2 is BLOCK. Any other exit is a **FAIL** of the check whatever it expected, and the failure line prints the exit code and the hook's stderr.
- Fixed before tagging (W1 of #103), so the matrix never counts a crashed hook as coverage.
- If tightening the helpers turns existing checks red, each red check is a finding. It is investigated and reported in the PR, never silenced by relaxing the helper.

## Acceptance criteria

- [ ] `check`, `check_in`, `flip` and any other helper that reads a hook's exit status apply exit 0 = ALLOW, exit 2 = BLOCK, anything else = FAIL.
- [ ] A self-test of the helper, in the manner of the suite's existing helper self-test: a fixture hook that exits 1, and one that exits 127, each fails an ALLOW expectation and a BLOCK expectation. The self-test asserts on the helper's result, not the fixture's exit code.
- [ ] The suite's header replaces the one-cause comment with the class.
- [ ] The self-test fails with the fix reverted.

## Testing decisions

Seam 1 plus a helper self-test. Literals throughout.

Blocks W1 of #103.

PR #111, opened 2026-09-15
Compare the worktree test as directories, and check it from every depth (#94)

### LINKED ISSUE #94: Linked-worktree detection compares paths as strings: a push from a subdirectory of the main checkout is permitted

## Problem

`no-git-push.sh` decides whether a command runs in a linked worktree by comparing `git rev-parse --git-dir` with `git rev-parse --git-common-dir` **as strings** (`no-git-push.sh:130-135` at `origin/dev-05` 750aace). At the root of a checkout git prints both relatively (`.git`, `.git`). Below the root, git 2.43 prints `--git-dir` as an absolute path and `--git-common-dir` as a relative one, so the two never compare equal and **the main checkout reads as a linked worktree**.

Measured in a throwaway repository on branch `feature-x`, piping `git push origin feature-x` into the committed hook:

| cwd | `--git-dir` | `--git-common-dir` | exit |
|---|---|---|---|
| repo root | `.git` | `.git` | **2** (refused) |
| `src/` | `/…/r/.git` | `../.git` | **0** (permitted) |
| `src/deep/` | `/…/r/.git` | `../../.git` | **0** (permitted) |

So an agent in the main checkout pushes any branch it likes, provided the command's working directory is below the repository root. This is the boundary's central rule (user stories 3 and 25 of #36; CONTEXT.md, *Worktree branch*: "the permission keys on where the command runs").

`no-work-on-stale-branch.sh:210-213` makes the same comparison and fails the other way: from a subdirectory of the main checkout it applies the worktree-lifecycle rules to the main checkout and can refuse an ordinary commit there.

**Why the suite is green.** `check-hooks.sh` runs every hook from `.claude/hooks` (the `check` helper, and `check_in` for its fixtures), and its context banner (around line 125) uses the same string comparison to decide which run happened. The main-checkout refusal at `no-git-push.sh:132-135` is never reached by a check: mutating it to `exit 0` leaves the suite green, from either context. Stories 2–4 are covered only when the suite happens to run in a linked worktree (95 failing mutants there, 0 from the main checkout).

kind: `defect-permitting` (push hook), `defect-refusing` (stale guard)

- **Q21 — the suite builds its own fixture repository**: a main checkout, a linked worktree made with `git worktree add`, and a subdirectory two levels deep in each. Every push-hook and stale-guard check runs in a named directory of that fixture, never in the invoker's context. The context-banner exception in the suite's header is removed: no expected verdict depends on where the suite is run. This is still seam 1 (a process, JSON on stdin, a known cwd).
- The keying stays *where the command runs*, never the branch name (CONTEXT.md, *Worktree branch*). The fix canonicalises both paths before comparing (for example `git rev-parse --path-format=absolute`, git ≥ 2.31, or resolving both through `cd … && pwd -P`); whichever is chosen, the header of each hook records why a string comparison was wrong.
- Both hooks are fixed together: one mechanism, one defect, two consumers. If the comparison moves into `lib/command-scan.sh`, it is covered by the load contract like every other `cs_*` function.

## Acceptance criteria

- [x] The fixture repository exists and every `no-git-push.sh` and `no-work-on-stale-branch.sh` check names its directory in it.
- [x] From the main checkout's root, `src/` and `src/deep/`: a push of the checked-out branch is **BLOCK** (three checks).
- [x] From the linked worktree's root, `src/` and `src/deep/`: a push naming its own branch is **ALLOW** (three checks), and a push of `main` or `dev-NN` is **BLOCK**.
- [x] From a subdirectory of the main checkout, the stale guard does not apply worktree-lifecycle rules: an ordinary commit is **ALLOW**.
- [x] Every new check fails with the fix reverted (record the mutation in the PR description).
- [x] The context banner is gone from `check-hooks.sh`, and the suite's header no longer describes an expectation that depends on context.
- [x] Each hook's header records the absolute/relative defect and the measurement above.

## Testing decisions

Seam 1 only. Expected verdicts are literals. Both directions for each location. No check derives its expectation by running the hook.

B1 blocks W1 of #103: tagging checks before they move into the fixture would tag them twice.

bgunyel:
Closes #94. Part of #36; one of the two blockers (with #98) of #104, W1 of #103.

## The defect

`no-git-push.sh` and `no-work-on-stale-branch.sh` decided "main checkout or linked worktree" by comparing `git rev-parse --git-dir` with `--git-common-dir` **as strings**. At the root of a checkout git 2.43 prints both relatively (`.git`, `.git`); below the root it prints `--git-dir` absolute and `--git-common-dir` relative (`/…/r/.git` against `../.git`), so the main checkout read as a linked worktree.

- `no-git-push.sh`, in the permitting direction: a push of the checked-out branch from the main checkout was exit 2 at the root and exit 0 from `src/` and `src/deep/`.
- `no-work-on-stale-branch.sh`, in the refusing direction: on the suite's lifecycle fixture, `git commit -m "wip"` in the main checkout was ALLOW at the root and BLOCK from `src/` and `src/deep/`.

## The fix

Both hooks resolve each path to the directory it names, `(cd -- "$1" && pwd -P)`, and compare those. `canonical_dir` is a copy in each hook rather than a `cs_*` function: the stale guard makes this comparison before it loads `lib/command-scan.sh`, so the common path of every git command costs no library load. The suite drives both copies.

`git rev-parse --path-format=absolute` gave the same answers at every depth measured, symlinked working directories included. It was rejected because git older than 2.31 echoes an unknown option as an output line, so both values would gain the same first line and differ in the second — the same defect again. Both hook headers record the measurement and this trade.

## The suite (Q21)

- The context banner, `OWN_BRANCH_PUSH` and `$CURRENT` are gone; the header no longer describes a context-dependent expectation.
- The suite builds a push fixture: a main checkout on `feature-x` (not `main`, where the reserved-branch rule would hide the defect), a linked worktree made with `git worktree add`, `src/deep/` in each, and a symlink to each. Every fixture directory, both branches and the `origin` remote are guarded, because a missing directory reads as ALLOW.
- **Every `no-git-push.sh` check names a directory in that fixture**: 180 former `check` calls, the four #68 "from main" checks and the eleven load-contract checks that used `$ON_DEV`.
- `no-work-on-stale-branch.sh` checks were already in named fixtures, one per lifecycle state; they cannot share one repository. The lifecycle fixture gains `src/deep/` in its main checkout and stale worktree, and a symlink.

New checks, as literals, both directions per location:

| where | command | verdict |
|---|---|---|
| main checkout: root, `src/`, `src/deep/` | `git push origin feature-x` | BLOCK ×3 |
| main checkout, `src/deep/` | same — refusal names the main checkout | says |
| main checkout: root, `src/`, `src/deep/` | `git fetch origin` | ALLOW ×3 |
| linked worktree: root, `src/`, `src/deep/` | `git push origin wt-branch` | ALLOW ×3 |
| linked worktree: root, `src/`, `src/deep/` | `git push origin main` / `dev-99` | BLOCK ×6 |
| through a symlink, `src/deep/` | main checkout push / worktree push | BLOCK / ALLOW |
| lifecycle main checkout: `src/`, `src/deep/`, symlink | `git commit -m "wip"` | ALLOW ×3 |
| stale worktree: `src/`, `src/deep/`, symlink | `git commit -m "wip"` | BLOCK ×3 |

## Mutations

Each run against the full suite on a fresh copy of the tree; green without a mutation (1033 checks).

| mutation | failing checks |
|---|---|
| `no-git-push.sh` back to the string comparison (the fix reverted) | 4 — main checkout `src/` and `src/deep/` BLOCK, the refusal message, the symlink BLOCK |
| `no-work-on-stale-branch.sh` back to the string comparison | 3 — main checkout `src/`, `src/deep/`, symlink ALLOW |
| `-P` dropped from `pwd`, `no-git-push.sh` | 1 — symlink BLOCK |
| `-P` dropped from `pwd`, `no-work-on-stale-branch.sh` | 1 — symlink ALLOW |
| main-checkout refusal `exit 2` → `exit 0` (the mutant #94 says survived) | 4 — main checkout BLOCK at root, `src/`, `src/deep/`, symlink |
| every directory read as the main checkout | 29 — every own-branch ALLOW, the new worktree ones included |

Before the fix, and before the symlink checks were added, the suite run from the main checkout and from a linked worktree failed the same five checks, so no verdict depends on where it is run.

## Trades, recorded

- **Not every new check fails with the fix reverted**, as the issue's acceptance criterion asks. At the root of a checkout, and anywhere in a linked worktree, the string comparison already answered correctly, so the root BLOCK, the `git fetch` ALLOWs, the nine worktree checks and the stale-worktree BLOCKs pass either way. The issue requires them anyway; they catch a fix that goes too far, as the last two mutations show.
- ~~**An unresolvable `--git-common-dir`** … no check covers it.~~ Closed in e773c6f after review: the push hook now refuses it with a message of its own, and a git shim reaches it for both hooks.
- ~~**`CDPATH`** … left unguarded.~~ Closed in e773c6f after review: `canonical_dir` empties CDPATH, which is checked by running the function itself against a decoy (not seam 1). The two copies of the function are also held identical as text.
- CONTEXT.md's *Worktree branch* entry names the comparison and sends the reader to the header for the mechanism; it is still accurate and is unchanged.


bgunyel:
## Review against #94

**Ready to merge, with one open question.** Six of the seven acceptance criteria are met. The seventh ("every new check fails with the fix reverted") can't be met as written, and the PR description says so.

### What I checked
- **The suite passes** on `34c9b19`: 1033 checks, exit 0, no failures.
- **It no longer depends on where it runs.** Run from the main checkout and from a linked worktree, the output was identical apart from temp-directory paths (Q21).
- **The mutation table.** Each mutation was run against the full suite:

| Mutation | Checks that failed | PR's claim |
|---|---|---|
| `no-git-push.sh` back to the string comparison | 4: main checkout `src/` and `src/deep/` BLOCK, the refusal message, the symlink BLOCK | 4 ✓ |
| `no-work-on-stale-branch.sh` back to the string comparison | 3: main checkout `src/`, `src/deep/` and symlink ALLOW | 3 ✓ |
| `-P` dropped from `pwd`, push hook | 1: symlink BLOCK | 1 ✓ |
| `-P` dropped from `pwd`, stale guard | 1: symlink ALLOW | 1 ✓ |
| Both hooks restored from `dev-05` | 7 | — |

Not re-run: the `exit 2` → `exit 0` mutation and "every directory read as the main checkout".

### Acceptance criteria
| # | Criterion | Verdict |
|---|---|---|
| 1 | The fixture exists; every push-hook and stale-guard check names its directory in it | ✅ with a disclosed exception. No push-hook check still uses `check` or `$ON_DEV`/`$ON_MAIN`. Some stale-guard checks still run in separate fixtures (`$NODEV`, `$WT_DIV`), because those lifecycle states can't share one repository. That's reasonable. |
| 2 | Main checkout root / `src/` / `src/deep/`: pushing the checked-out branch is BLOCK | ✅ three checks, plus a `says` on which refusal fired |
| 3 | Linked worktree root / `src/` / `src/deep/`: own branch ALLOW; `main` and `dev-NN` BLOCK | ✅ 3 ALLOW + 6 BLOCK |
| 4 | Stale guard from a subdirectory of the main checkout: an ordinary commit is ALLOW | ✅ `src/`, `src/deep/`, and through the symlink |
| 5 | Every new check fails with the fix reverted | ❌ **not met as written** (see below) |
| 6 | The context banner is gone; the header describes no context-dependent expectation | ✅ `OWN_BRANCH_PUSH` and the banner are gone. The one remaining string comparison of the two paths only prints a note about git's behaviour; it asserts nothing. |
| 7 | Each hook's header records the absolute/relative defect and the measurement | ✅ both headers, including why `--path-format=absolute` was rejected |

**On criterion 5:** with both hooks reverted, 7 of the roughly 24 new checks fail. The rest pass either way: the root BLOCK, the three `fetch` ALLOWs, the nine worktree checks, the worktree symlink ALLOW and the stale-worktree BLOCKs. The string comparison already answered those locations correctly. Criterion 3 requires those checks, so criteria 3 and 5 contradict each other. The PR's argument holds: those checks catch a fix that goes too far, and the mutation run backs that up. I suggest rewording criterion 5 in #94 to "every check the defect can affect fails with the fix reverted", rather than leaving it unticked.

### Minor points (none blocking)
1. **The refusal message can be wrong.** If `--git-common-dir` doesn't resolve, the push hook still refuses, which is the right default. But the message says "This is the main checkout, not a linked worktree", which isn't true on that path. No check reaches it.
2. **`CDPATH`.** The reasoning in the PR holds: git only prints `.git` (the form `CDPATH` affects) at the root, where both sides resolve alike. Still, `CDPATH= cd -- "$1"` would close it at no cost and remove a paragraph of justification.
3. **Two copies of `canonical_dir`.** Only the behavioural checks keep them in step; nothing checks that the text is the same. That's acceptable given the reason (the stale guard compares before it loads `lib/command-scan.sh`), but the copies could drift in ways the suite doesn't exercise.

bgunyel:
## The three minor points, addressed in e773c6f

Each point now has a check that fails without its fix. The suite passes with 1046 checks (up from 1033).

### 1. The refusal message when a path can't be resolved
`no-git-push.sh` now handles an unresolvable path separately from the main checkout:

> The repository directory git reports here could not be resolved, so whether this runs in a linked worktree cannot be judged from here.

The main-checkout message is used only when both paths resolve and are equal. The stale guard still abstains on this path.

No real repository reaches this case: if git can't find its common directory, it can't find the repository either. So the suite puts a **git shim** first on `PATH`. The shim answers `--git-common-dir` with a directory that doesn't exist and passes every other call to the real git. The suite checks that the shim fakes that one answer and nothing else. The shim's checks run where getting it wrong would change the verdict:
- **linked worktree:** a push of its own branch is BLOCK. The refusal says "could not be resolved", and does not say "This is the main checkout".
- **stale worktree:** a commit is ALLOW, meaning the guard abstains. Treating the path as a worktree would refuse it.
- **outside any repository:** a push is BLOCK with the same new message. Before this change it got the main-checkout message too.

### 2. `CDPATH`
`canonical_dir` now runs `(CDPATH= cd -- "$1" … && pwd -P)`.

No check that runs a hook can see this, for the reason the PR description gave: both halves are `.git` at the root, and below the root git never prints a path that `CDPATH` consults. So the function is checked on its own. The suite extracts it from each hook, evaluates it in a subshell with `CDPATH` pointing at a decoy that holds a `.git`, and expects the fixture's own `.git`.

The expected path comes from `readlink -f`, not `pwd -P`, so the expectation doesn't go through the mechanism under test. A fixture check first confirms that a plain `cd` really does follow the decoy there. **This is not seam 1**, and the suite says so where the check is written.

### 3. The two copies of `canonical_dir`
The copies are now held together the same way as the dev-branch derivation copied between the stale guard and `report-stale-branches.sh`:
- each hook defines the function exactly once;
- the two texts are identical;
- each matches the text pinned in the suite.

A rewording of one copy that keeps its behaviour, which is exactly what the behaviour checks can't catch, now fails the suite.

### Mutations
Each was run against the full suite on a fresh copy of the tree.

| Mutation | Failing checks |
|---|---|
| Unresolved-path check dropped, push hook | 2: shim BLOCK, and the "could not be resolved" message check |
| Unresolved-path check dropped, stale guard | 1: shim ALLOW |
| Unresolved case given the main-checkout message again | 3: the two "says" checks and the "must not say" check |
| `CDPATH=` removed, push hook | 3: copies identical, push hook text matches the pin, push hook `CDPATH` check |
| `CDPATH=` removed, stale guard | 3: copies identical, stale guard text matches the pin, stale guard `CDPATH` check |
| `[ -n "$1" ]` rewritten as `test -n "$1"` in one copy | 2: copies identical, that copy matches the pin |

### Also changed
- **PR description:** the two "Trades, recorded" bullets these points closed are struck through and point to e773c6f.
- **Hook headers:** both are updated for the new message, the shim and `CDPATH`.

### Not done
- **Criterion 5 of #94:** your suggested rewording of the acceptance criterion is left to you.
- **`make upgrade-safe`:** not run for this change.

PR #112, opened 2026-09-15
Read every hook's tool call through one reader that refuses what it cannot read (#95)

### LINKED ISSUE #95: Every hook permits when jq is absent or its input is malformed

## Problem

Every hook reads its input with its own `jq` call (`no-git-push.sh:78`, `no-pr-decisions.sh:113`, `no-commit-to-main.sh:86`, `no-work-on-stale-branch.sh:198`, `pytest-via-uv-group.sh:84`, `alembic-via-uv-group.sh:37`, `append-only-docs.sh:24`, `append-only-docs-edit.sh:29`, at `origin/dev-05` 750aace). When that read fails, `COMMAND` is empty and every hook exits 0.

Measured against the committed hooks, with `git push --force origin main` and `gh pr merge 5` as the commands (both refused, exit 2, in a normal environment):

| condition | `no-git-push.sh` | `no-pr-decisions.sh` | `no-commit-to-main.sh` |
|---|---|---|---|
| `jq` absent from `PATH` | **0** | **0** | **0** |
| stdin not JSON | **0** | – | – |
| empty stdin | **0** | – | – |
| `{"tool_name":"Bash","tool_input":{}}` (no command) | **0** | – | – |

`jq` absent is #84's finding again in a different place. #84 made a hook that cannot load `lib/command-scan.sh` refuse rather than permit; it did not cover the tool every hook needs before it reaches the library. No check pins any row of this table.

kind: `defect-permitting`

- **Q19 — fail closed** when `jq` is absent, when stdin is not valid JSON, or when the field the hook reads (`tool_input.command`; `tool_input.file_path` for the Edit hook) is missing or not a string. **Fail open only for an empty command string**: there is nothing to run. The reasoning is CLAUDE.md's left-open item 3: a blocked command is visible and one edit away, a silently permitted push is neither. A change to the harness's payload format then shows up as every Bash call refused, not as every guard switched off without a word.
- **Q27 — every hook, enforced once.** The seven Bash hooks and the Edit hook. The input read moves into one shared reader in `lib/command-scan.sh`, beside the load contract #84 established, so eight copies cannot disagree. `append-only-docs-edit.sh` does not source the library today (its line 49 says why); it either starts sourcing it for the reader, or the PR records why it keeps its own copy and a check pins that copy's refusals identically.
- **The trade is recorded, not just the fix.** An environment without `jq` will refuse every Bash command, including `pytest` and `alembic` invocations the convention hooks would otherwise permit. The refusal message names the cause ("jq is not on PATH") so the fix is one install away. The trade goes in the library's header and in a check.

## Acceptance criteria

- [ ] With `jq` absent from `PATH` (fixture: a directory of symlinks to the system tools minus `jq`), each of the eight hooks exits **2** with a message naming `jq`.
- [ ] Invalid JSON, empty stdin, and a missing or non-string command field: each Bash hook exits **2**; the Edit hook the same for `file_path`.
- [ ] An empty command string (`"command":""`): each Bash hook exits **0**.
- [ ] Normal payloads are unchanged: the full existing suite stays green.
- [ ] Every new check fails with the fix reverted.
- [ ] CLAUDE.md or the library header states the fail direction for each condition in one place.

## Testing decisions

Seam 1. Literal verdicts. The `jq`-less `PATH` is a fixture built by the suite, not an assumption about the machine. After B5 lands, "permit" means exit exactly 0.

Blocks W5 of #103.

bgunyel:
Closes #95. Part of #36; unblocks W5 of #103.

## The defect

Every hook read its tool call with its own `jq -r`. When that read failed, the command was empty and the hook exited 0. At dev-05 e8c132f, all eight hooks permitted every one of these conditions:

- `jq` off PATH
- stdin not JSON, cut off, empty, or followed by text
- a `command` or `file_path` that was missing or not a string

## The fix

`cs_tool_input <field>` in `lib/command-scan.sh` is now the only read, and all eight hooks call it (#103 Q27). It refuses (exit 2) in three cases:

- `jq` is not on PATH. The message names `jq`, so the fix is one install away.
- stdin is not exactly one JSON value.
- `tool_input.<field>` is missing or not a string.

It permits only a string. An empty command string reads as empty, and the hooks permit it: there is nothing to run (#103 Q19).

Two details matter:

- **It reads all of stdin before printing.** `jq` exits 0 on empty input, and it prints the first value before failing on trailing text. So neither its exit status nor its output is enough on its own.
- **It checks the field's type inside `jq`.** `jq -r` prints a JSON null as the string `null`.

The fail direction for every condition is stated once, under **THE INPUT READ** in the library header. Every hook checks for the reader in its load guard. The two append-only hooks now source the library for this reader alone, so the Edit hook keeps no copy of its own.

## Trades, recorded in the code and pinned by checks

- **Without `jq`, everything is refused.** That covers every Bash command and every Edit or Write, including the `uv run` forms of pytest and alembic that the convention hooks exist to permit.
- **`no-work-on-stale-branch.sh` now loads the library before reading input.** So a missing `lib/` also refuses on a branch carrying work, where it used to permit. This is recorded as a `flip`, ALLOW → BLOCK. With `lib/` gone, every other Bash hook already refused every command.
- **An empty `file_path` is still permitted.** It names no file. #95 did not decide this case, so this PR does, and pins it.

## Checks

- **New section, `issue #95`, at the foot of `check-hooks.sh`.**
  - New helpers `feed` and `feed_says` hand a hook raw stdin and read the exit code exactly: 0 is ALLOW, 2 is BLOCK, anything else FAILs (the #98 rule, applied here).
  - The `jq`-less PATH is a fixture the suite builds: symlinks to every tool on PATH, with a twin that includes `jq` as its control.
- **Load contract.** It now covers eight consumers, and it tests `cs_tool_input` with the function missing for each one.
- **Derived checks.**
  - The hooks checked are exactly the ones `settings.json` registers, for Bash, Edit and Write.
  - No hook calls `jq` itself.

**Against dev-05:** every BLOCK check and every message check fails. The ALLOW checks pass there by design; they catch a fix that refuses too much, and each one names the over-fix it catches.

**Mutation-checked in scratch copies: 14 mutants, all killed.** Two of them survived the first pass:

- **The reader's object-type tests.** They were dead code, because indexing a non-object is already a `jq` error, so they were removed.
- **The "exactly one value" test.** A check for two back-to-back JSON values was added.

## Evidence

```
bash .claude/hooks/check-hooks.sh
ALL CHECKS PASSED   (1224 ok, 0 FAIL)
```

Hook latency is unchanged, roughly 10–80 ms per call, measured against dev-05.

## Left open

The "no hook calls `jq` itself" check strips comments the same way `armed` does. That also cuts a line at a `${x#y}` expansion, so a `jq` call after one on the same line would not be seen by that text check. The comment above the check says so.

`make upgrade-safe` has not been run on this branch. No dependency changed, but CLAUDE.md requires it before merge.


bgunyel:
## Review against #95's acceptance criteria

Reviewed by the assistant at head `77f50f2`. **No defect found in the fix. Not ready to merge until the three items under _Before merge_ are settled.**

### What was run

- **Suite on the PR head:** `bash .claude/hooks/check-hooks.sh` gave 1224 ok, 0 FAIL.
- **Suite on the PR merged with current `dev-05`** (`d71ab1c`, which includes #111): the merge is clean with no conflicts, and the suite gave **1261 ok, 0 FAIL**. The "no hook calls `jq` itself" check still passes, so #111 added no second input read.
- **Revert test:** this PR's `check-hooks.sh` against the nine hook and library files from `dev-05` `e8c132f`.
- **Extra inputs:**
  - A trailing newline, CRLF and a leading BOM are all read correctly.
  - A 5 MB Write payload is read in 96 ms.
  - A normal `git push --force origin main`, a `gh pr merge 5`, and an edit to an existing `docs/dev-log/` entry are all still refused.
- **Latency:** within 0–2 ms of `dev-05` on the four hooks timed.

### Acceptance criteria

| # | Criterion | Verdict |
|---|---|---|
| 1 | `jq` absent → all eight hooks exit 2, naming `jq` | ✅ All eight. The PATH fixture is built by the suite, with a with-jq twin as the control. |
| 2 | Invalid JSON, empty stdin, missing or non-string field → exit 2 (`file_path` for the Edit hook) | ✅ All pass here and all fail on `dev-05`. Also covers cut-off input, trailing text, two values, `null` and `false`. |
| 3 | `"command":""` → each Bash hook exits 0 | ✅ |
| 4 | Normal payloads unchanged, existing suite green | ✅ One existing #84 check changes its verdict; see _Before merge_ 2. |
| 5 | Every new check fails with the fix reverted | ⚠️ Met for every BLOCK check and every message check: 150 fail in the #95 section. The 26 ALLOW checks and the 3 `settings.json` derivation checks pass on `dev-05`. A check that expects ALLOW cannot go red when a fix that only adds refusals is removed, and the derivations read `settings.json`, not hook behaviour. The PR body says this. |
| 6 | Fail direction stated in one place | ✅ THE INPUT READ in `lib/command-scan.sh`, with THE TRADE beside it. Both are pinned. |

### Before merge

1. **`make upgrade-safe` has not been run.** CLAUDE.md requires it before a PR closes. No dependency changed, so a pass is expected, but it has not been run.
2. **The #84 check reversed by this PR needs a decision.** With `lib/` missing, `no-work-on-stale-branch.sh` now refuses on a branch carrying work, where #84 pinned ALLOW. The assistant recommends accepting it: with `lib/` gone, every other Bash hook already refuses every command.
3. **Empty `file_path` → ALLOW needs a decision.** #95 left this case open. The assistant recommends accepting it: an empty path names no file, and it is the existing behaviour.

### Minor

- **The revert result is hard to reproduce.** Run against the `dev-05` hooks, this suite exits 1 at the `mk_halflib` fixture guard ("did not rename cs_tool_input away") before it reaches the #95 section. The 150 failures above were counted with that guard disabled in a scratch copy. The suite still goes red, so criterion 5 is not at risk, but a line in the PR body on how "against dev-05" was measured would help the next reviewer.
- **The PR body is out of date.** It quotes 1224 checks against `e8c132f`; on current `dev-05` the count is 1261.
- **Criterion 5 in #95** could read "every BLOCK check and every message check fails with the fix reverted", so it matches what can be checked.
- **`requirements.md` (#104) is not on `dev-05` yet**, so no `GH-95` entry is owed now. It will be needed when #104 lands.

bgunyel:
## Review points addressed

Answering the review above, item by item. The PR head is now `4025413`.

### Before merge

1. **`make upgrade-safe`: not run.** Bertan asked for it to be left out of this round, so it is still outstanding before merge.
2. **#84 check reversed: accepted by Bertan.** With `lib/` missing, `no-work-on-stale-branch.sh` now refuses on a branch carrying work. The PR description records the acceptance; the `flip` in the suite already pinned the new verdict.
3. **Empty `file_path` → ALLOW: accepted by Bertan.** Recorded in the description. The hook and a check already pinned it.

### Minor

- **The revert result was hard to reproduce.** Re-measured by the assistant on the merge with `dev-05`, and the description now says how.
  - **Setup:** this branch's `check-hooks.sh`, in a scratch copy, with the eight hooks and `lib/command-scan.sh` taken from `origin/dev-05`.
  - **The one change to the suite:** the `mk_halflib "$hook" cs_tool_input` loop is replaced by `:`. It stops the suite early on an old library, which has no `cs_tool_input` to rename away.
  - **Result:** 150 of the 179 checks in the #95 section fail and 29 pass, the same numbers the review got.
  - **What fails:** 131 BLOCK checks, 8 message checks, 8 "does not call `jq` itself" checks, and 3 text pins.
  - **What passes:** 26 ALLOW checks and 3 `settings.json` derivations.
- **The description's check count was stale.** `dev-05` (`d71ab1c`, which includes #94 via #111) is merged in as `4025413`. The merge was clean. In `no-work-on-stale-branch.sh` the input read still runs before #94's `canonical_dir` git reads. The suite passes at `4025413` with **1261 ok, 0 FAIL**, and the description now quotes that.
- **#95's criterion 5 reworded.** It now reads "Every new BLOCK check and every new message check fails with the fix reverted", with a note in the issue on when and why. That was the only line of the issue changed.
- **`requirements.md` (#104):** nothing is owed until it lands on `dev-05`.

PR #116, opened 2026-09-15
Read a hook's exit as 0, 2 or FAIL, so a crashed hook is no longer a pass (#98)

### LINKED ISSUE #98: check-hooks.sh reads any exit other than 2 as ALLOW: a crashed hook passes every permitting check

bgunyel:
Closes #98. Part of #36; W1 of #103.

## What was wrong

Every helper in `check-hooks.sh` that ran a hook read its exit status as one bit: 2 was BLOCK, anything else ALLOW. A hook that crashed or was never found passed every ALLOW-expecting check it was given. `says_not` did not read the status at all, so a crashed hook passed it whenever its stderr lacked the fragment. That hole is outside the issue's acceptance criteria but inside its problem statement, and it was found by review of this change.

## What changed

- **`verdict`** reads the status once for `check`, `check_in` (and so `flip`) and `check_file`: exit 0 is ALLOW, exit 2 is BLOCK, and anything else FAILs whatever was expected. The failure line carries `exit=<status>` and the hook's stderr.
- **`says` / `says_not`** FAIL unless the hook refused with exit 2. Every call site of both asserts on a refusal.
- **Comments.** The class is written once, above `verdict`, and the one-cause comments beside the fixture guards point there, including the push-fixture guard that #94 added.
- **Self-test** at the foot of the suite. It drives all six helpers with fixture hooks that exit 0, 2, 1 and 127, and asserts on the `FAILED` each helper leaves, not on the fixture's exit. Each fixture leaves a marker proving it ran. The set of helpers that read an exit status is derived from the file and compared with the literal list driven.

## Findings from tightening the helpers

**None.** No existing check turned red. That held on `dev-05` before #94 (1052 ok, 0 FAIL) and holds after merging #94 (1089 ok, 0 FAIL). Every hook the suite drives exits only 0 or 2 on the paths it drives.

## Evidence

Mutation-checked in scratch copies. The first three rows were run against the final self-test. The rest were run against its first version, before `says`/`says_not` joined it; the cases those rows exercise are unchanged since.

| Mutant | #98 checks that go red |
|---|---|
| `verdict` reverted to the one-bit reading | 16 (ALLOW expectation + failure line, for exit 1 and 127, in each of the four `check*` helpers) |
| exit-2 requirement removed from `says`/`says_not` | 8 |
| `check_in` leaking the hook's stderr instead of printing it | 4 (`check_in` and `flip` failure lines) |
| every nonzero exit read as BLOCK | 8 (the BLOCK-expecting crash cases) |
| `check_file` left on one bit | 4 |
| failure line without stderr | 8 |
| `drive_helper` losing its `cd` | every `check`/`check_file` case reports "the fixture never ran" |
| a fifth helper reading `rc=$?` | the derivation |

The crash cases that expect BLOCK do not fail on a revert, because the one-bit reading failed those too. They are there for the opposite wrong reading, row 4.

## The trade, and a correction to the issue

**A bash syntax error exits 2, not 1** as the issue says. This was measured. So a hook that does not parse still reads as BLOCK, and passes every BLOCK-expecting `check` against it. This PR does not close that case. It is recorded above `verdict`. `says` is what separates a refusal from a parse failure, where a check has one beside it.

A second limit of the derivation: it finds status readers spelled `rc=$?`. A helper written as `if "$hook"; then` would go unfound, and the comment says so.


bgunyel:
## Review against #98

**All four acceptance criteria are met.** I checked each one by running it. The findings below don't block merging.

Run at PR head `5467cd8` (up to date with `dev-05`), bash 5.2.21: **1089 ok, 0 FAIL, exit 0.**

| # | Criterion | | Evidence |
|---|---|---|---|
| 1 | `check`, `check_in`, `flip` and any other status-reading helper: 0 = ALLOW, 2 = BLOCK, else FAIL | ✅ | `$?` is read in exactly five places (`check`, `check_in`, `check_file`, `says`, `says_not`). The first three go through `verdict`, and `flip` goes through `check_in`. No hook is run outside these helpers. |
| 2 | Self-test: exit-1 and exit-127 fixtures each fail an ALLOW and a BLOCK expectation, asserting on the helper's result | ✅ | 43 new checks. They assert on the subshell's `FAILED`, not the fixture's exit code. A marker shows each fixture ran, and exit-0/2 control cases are included. |
| 3 | Header replaces the one-cause comment with the class | ✅ | The class is written above `verdict`, the header points to it, and all four one-cause guard comments now point to `verdict`. |
| 4 | Self-test fails with the fix reverted | ✅ | Mutants run in scratch copies (below). |

**Mutants:**
- **Full revert** (one-bit `verdict`, no exit-2 gate in `says`/`says_not`): 24 FAIL, all in the #98 self-test.
- **`verdict` reverted alone:** 16 FAIL.
- **Exit-2 gate removed alone:** 8 FAIL.

These match the table in the description. The correction to the issue also holds: a bash syntax error exits 2, which I measured here too.

### Finding: the helper list found in the file isn't compared with the helpers the test runs

The description says the found set is "compared with the literal list driven". In the code, `STATUS_READERS` is compared only with the literal string passed to `tok`. There are three separate hard-coded lists (the `check check_in flip check_file` loop, the `says says_not` loop, and the `tok` string), and nothing ties the `tok` string to the two loops.

**Repro:** I added a helper that reads `rc=$?` with the old one-bit logic:
```bash
check_env() {  # check_env <script> <want> <label>
  local rc err
  err=$(./"$1" 2>&1 >/dev/null)
  rc=$?
  case "$rc" in 2) got=BLOCK ;; *) got=ALLOW ;; esac
}
```
I added its name only to the `tok` literal (`'check check_env check_file check_in says says_not '`). The suite stayed green: **1089 ok, 0 FAIL.** A new reader goes red, gets "fixed" by editing one string, and is never tested. That is the kind of silent pass #98 is meant to remove.

**Suggested fix:** name the driven lists once and check that every found reader is in them:
```bash
DRIVEN_VERDICT='check check_in flip check_file'
DRIVEN_MESSAGE='says says_not'
for helper in $DRIVEN_VERDICT; do …   # existing body
for helper in $DRIVEN_MESSAGE; do …   # existing body
for reader in $STATUS_READERS; do
  present "$reader is driven by the #98 self-test" "$reader" "$DRIVEN_VERDICT $DRIVEN_MESSAGE"
done
```
The other option is to change the description to say what the comparison actually covers.

### Nits
1. **The new header paragraph doesn't say what the class is.** It says one exists and points to `verdict`. Adding "exit 0 is ALLOW, 2 is BLOCK, anything else FAILs" would let the header stand on its own.
2. **The helper search has an unstated limit.** The pattern `^[a-z_]+\(\) *\{` misses names with digits or capitals, and indented or `function`-keyword definitions. Those would go unnoticed the same way as the `if "$hook"; then` case the comment already names, so they could go in the same sentence.

bgunyel:
Thanks. The finding is right, and it is fixed in 8a5e405.

**The finding: derived readers are now tied to the loops that drive them.**
- The driven helpers are named once, in `DRIVEN_VERDICT` and `DRIVEN_MESSAGE`, and each loop runs off its list.
- Every derived reader must be a member of one of those lists, checked with `present` as you suggested. The `tok` literal is gone, so there is no third list to edit.
- An empty derivation now FAILs. Otherwise the membership loop would pass by asking nothing, which the suggested fix would also have allowed.

Your repro, run against a scratch copy: the one-bit `check_env` reader goes red with `FAIL derived check_env reads a hook exit status, and the #98 self-test drives it`.

**Nit 1.** The header paragraph now states the rule: exit 0 is ALLOW, exit 2 is BLOCK, anything else FAILs whatever was expected.

**Nit 2.** I widened the pattern rather than only documenting it. Names with capitals or digits are found now, and so is the `function` keyword with or without parens. Both were mutation-checked: `Check_Env2() {` and `function check_env {` each go red on the membership check. What it still misses is named in the comment:
- a reader spelled other than `rc=$?`;
- a function defined indented, whose body has no `}` at the start of a line to end it.

Suite at 8a5e405: 1093 ok, 0 FAIL. The four extra ok lines come from one `tok` replaced by five membership checks. The description's evidence table now has rows 9–12 for these mutants.

